PR #48, opened 2019-06-25
Support compiled TypeScript

### LINKED ISSUE #7: Check babel macro can work with typescript
- [x] shift-babel-macro runs when imported
 - [x] AST transformation can result in compilable typescript code (arik note: may not be needed if we support compiled typescript)

arikmaor:
Fix #7

### DIFF
--- a/babel-macro/src/macro.ts
+++ b/babel-macro/src/macro.ts
@@ -27,12 +27,17 @@ interface MacrosBabel {
 function findExportedMethods(ast: babelTypes.File, { types: t }: MacrosBabel): string[] {
   const body = ast.program.body;
   // support ExportDefaultDeclaration (ExportAllDeclaration too for re-exporting ?)
-  const namedExports = body.filter((e): e is babelTypes.ExportNamedDeclaration => t.isExportNamedDeclaration(e));
-  const exposedExports = namedExports.filter(
-    (e) => e.leadingComments && e.leadingComments.some((comment) => /@expose/.test(comment.value)));
-  const functions = exposedExports.map((e) => e.declaration).filter(
-    (e): e is babelTypes.FunctionDeclaration => t.isFunctionDeclaration(e));
-  return functions.filter((f) => !!f.id).map((f) => f.id!.name);
+  const exposedStatements = body.filter((e) => e.leadingComments && e.leadingComments.some((comment) => /@expose/.test(comment.value)));
+  const exposedFunctions = exposedStatements.reduce((ret: babelTypes.FunctionDeclaration[], e) => {
+    if (t.isFunctionDeclaration(e)) {
+      ret.push(e);
+    } else if (t.isExportNamedDeclaration(e) && t.isFunctionDeclaration(e.declaration)) {
+      ret.push(e.declaration)
+    }
+    return ret;
+  }, []);
+
+  return exposedFunctions.filter((f) => !!f.id).map((f) => f.id!.name);
 }
 
 function shiftMacro({ state, babel }: { state: MacrosPluginPass, babel: MacrosBabel }) {
--- a/babel-macro/src/macro.ts
+++ b/babel-macro/src/macro.ts
@@ -29,7 +29,7 @@ function getFunctionName(e: babelTypes.FunctionDeclaration, t: typeof babelTypes
 }
 
 function verifyExportedMember(ast: babelTypes.File, t: typeof babelTypes, funcName: string): void {
-  const body = ast.program.body;
+  const { body } = ast.program;
   const isFunctionExported = body.some((e) =>
     t.isExpressionStatement(e) &&
     t.isAssignmentExpression(e.expression) &&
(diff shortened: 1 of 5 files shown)


babel-macro/src/macro.ts
```diff
@@ -24,15 +24,48 @@ interface MacrosBabel {
   types: typeof babelTypes;
 }
 
-function findExportedMethods(ast: babelTypes.File, { types: t }: MacrosBabel): string[] {
+function getFunctionName(e: babelTypes.FunctionDeclaration, t: typeof babelTypes): string | undefined {
+  return t.isIdentifier(e.id) ? e.id.name : undefined;
+}
+
+function verifyExportedMember(ast: babelTypes.File, t: typeof babelTypes, funcName: string): void {
+  const body = ast.program.body;
```
michaeladda:
```suggestion
  const { program: { body } } = ast;
```

babel-macro/src/macro.ts
```diff
@@ -24,15 +24,48 @@ interface MacrosBabel {
   types: typeof babelTypes;
 }
 
-function findExportedMethods(ast: babelTypes.File, { types: t }: MacrosBabel): string[] {
+function getFunctionName(e: babelTypes.FunctionDeclaration, t: typeof babelTypes): string | undefined {
+  return t.isIdentifier(e.id) ? e.id.name : undefined;
+}
+
+function verifyExportedMember(ast: babelTypes.File, t: typeof babelTypes, funcName: string): void {
```
michaeladda: maybe assetExportedMember (since it throws otherwise)

babel-macro/src/macro.ts
```diff
@@ -24,15 +24,48 @@ interface MacrosBabel {
...
+  const body = ast.program.body;
+  const isFunctionExported = body.some((e) =>
+    t.isExpressionStatement(e) &&
+    t.isAssignmentExpression(e.expression) &&
+    e.expression.operator === '=' &&
+    t.isMemberExpression(e.expression.left) &&
+    t.isIdentifier(e.expression.left.object) &&
+    e.expression.left.object.name === 'exports' &&
+    t.isIdentifier(e.expression.left.property) &&
+    e.expression.left.property.name === funcName &&
```
michaeladda: iiuc this only specifically matches `exports.funcName = funcName`.
If so, please add a comment. If not please add a comment with the actual example.
Also, please document the fact it is the current typescript generated pattern.

babel-macro/src/macro.ts
```diff
@@ -24,15 +24,48 @@ interface MacrosBabel {
...
-  const exposedExports = namedExports.filter(
-    (e) => e.leadingComments && e.leadingComments.some((comment) => /@expose/.test(comment.value)));
-  const functions = exposedExports.map((e) => e.declaration).filter(
-    (e): e is babelTypes.FunctionDeclaration => t.isFunctionDeclaration(e));
-  return functions.filter((f) => !!f.id).map((f) => f.id!.name);
+  const exposedStatements = body.filter((e) => e.leadingComments && e.leadingComments.some((comment) => /@expose/.test(comment.value)));
+  return exposedStatements.reduce((ret: string[], e) => {
+    if (t.isFunctionDeclaration(e)) {
+      const funcName = getFunctionName(e, t);
+      if (funcName && importedNames.includes(funcName)) {
```
michaeladda: why limit this only to imported names (`importedNames.includes(funcName)`)? So we can throw during import?
if so, why doesn't the function name in line 62 gets the same treatment?

babel-macro/src/test/macro.spec.ts
```diff
@@ -87,5 +94,20 @@ pluginTester({
         });
       `,
     },
+    "Supports TypeScript output": {
```
michaeladda: we need to add all the javascript tests above in typescript too

babel-macro/src/macro.ts
arikmaor: changed

babel-macro/src/macro.ts
arikmaor: Comment added

babel-macro/src/macro.ts
arikmaor: given this `backend` file:
```js
// @expose
export function foo() {
}

// @expose
function internalFunctionWithObsoleteExposeComment() {
}
```

And N frontend files like this:
```js
import { foo } from '../backend/myFile'

...
```

I don't want all the frontend files compilations to fail because of an unrelated compilation error in a different file that would not cause anything wrong in the front-end code.
At most, I want this error only in the backend file compilation.

When importing `foo` I do want this error because this function is actually called from the frontend code which I'm currently compiling, so it wouldn't work if the function is not exported and exposed

babel-macro/src/test/macro.spec.ts
michaeladda: after f2f it is obvious this is not required, as the output should be tested as vanilla js

PR #51, opened 2019-06-26
Add Circle CI integration

### LINKED ISSUE #31: Add CI integration

bergundy:
Closes #31

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,40 @@
+version: 2
+jobs:
+  build_and_test:
+    docker:
+      - image: circleci/node:10-browsers
+    steps:
+      - checkout
+      - run:
+          name: Check for missing change logs
+          command: |
+            git fetch origin master:refs/remotes/origin/master -a
+            node common/scripts/install-run-rush.js change --verify
+      - restore_cache:
+          key: dependency-cache-{{ checksum "common/config/rush/shrinkwrap.yaml" }}
+      - run:
+          name: Install dependencies
+          command: node common/scripts/install-run-rush.js install
+      - save_cache:
+          key: dependency-cache-{{ checksum "common/config/rush/shrinkwrap.yaml" }}
+          paths:
+            - common/temp
+      - run:
+          name: Build source
+          command: node common/scripts/install-run-rush.js rebuild --verbose
+      # - run:
+      #     name: Run lint
+      #     command: npm run lint
+      - run:
+          name: Run tests
+          command: |
+            for f in $(find . -maxdepth 2 -name package.json); do
+              project=$(dirname $f)
+              echo Testing $project
+              (cd $project && npm test) || exit 1
+            done
+workflows:
+  version: 2
+  build_and_test:
+    jobs:
+      - build_and_test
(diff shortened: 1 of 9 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,40 @@
...
+          name: Install dependencies
+          command: node common/scripts/install-run-rush.js install
+      - save_cache:
+          key: dependency-cache-{{ checksum "common/config/rush/shrinkwrap.yaml" }}
+          paths:
+            - common/temp
+      - run:
+          name: Build source
+          command: node common/scripts/install-run-rush.js rebuild --verbose
+      # - run:
```
arikmaor: Is this commented out on purpose?

babel-macro/package.json
```diff
@@ -7,7 +7,7 @@
   ],
   "description": "",
   "scripts": {
-    "build": "rm -rf dist && tsc && cp -R 'src/test/backend/' 'dist/test/backend'",
+    "build": "rm -rf dist && tsc && cp -r src/test/backend/ dist/test/",
```
arikmaor: From `man cp`
```
COMPATIBILITY
	Historic versions of the cp utility had a -r option.
	This implementation supports that option; however, its use is strongly discouraged, as it does not correctly copy special files, symbolic links, or fifo's.
```

babel-macro/src/test/macro.spec.ts
```diff
@@ -7,7 +7,7 @@ process.chdir(__dirname);
 pluginTester({
   plugin,
   babelOptions: {
-    filename: join(__filename, 'tested-file.js'),
+    filename: join(__dirname, 'babel-output', 'tested-file.js'),
```
arikmaor: The tests are passing after this change?

.circleci/config.yml
bergundy: Yes, I'll add a task to enable linting on all repos.

babel-macro/package.json
bergundy: It's not in the linux man page http://man7.org/linux/man-pages/man1/cp.1.html

babel-macro/src/test/macro.spec.ts
bergundy: Yes, you can see the status on the PR.

babel-macro/package.json
arikmaor: So you're saying that you *want to stop* supporting development on mac? (as you are specifically changing from the cross-platform option to the linux only option and this is the only actual change that is happening here)

I don't want to block, so I'm approving but from now on I'll take under-consideration that building on our machines is not necessary nor supporting developers who wants to contribute but don't know how to use docker...

vogre: I think the changes between BSD and Linux can be ignored, and something like `cp -a` would be unfamiliar.

bergundy: Just note that this still works on Mac..

arikmaor: Current state: using -R
Proposed change: using -r
difference in linux - none, they both work the same
difference in osx - going from recommended to deprected
difference in windows - none, they both don't work

So this change as I see it, is only to stick it to apple.
And it's an explicit change just for that, it causes no other change.

PR #59, opened 2019-06-30
Use a constant port for the local-proxy 

### LINKED ISSUE #36: Backend non response after code change
Steps (not sure if repeatable):

1. Run app
2. Change backend code
3. Change frontend code
4. Change backend code

Result:

Screen Shot 2019-06-20 at 12 27 09

vogre:
Opening a Server in the parent process attaches a listener to the server
'accept', this results in a possibility of the child process never
receiving some of the requests.

Closing the server in the parent process keeps the server open in the
child process, but prevents from passing the already closed server to
new child processes.

This limits the functionality to use only one port for development
(multiple npm run start will fail), but fixes #36 unpredictable
behavior.


local-proxy/src/index.ts
```diff
@@ -31,28 +23,24 @@ export async function startProxy(rootDir: string, localToken: string) {
...
   });
 
-  let handle: any;
-
   nodemon.on('quit', () => {
     process.exit();
-  }).on('start', (child) => {
-    handle = child;
+  }).on('start', (_child) => {
+    // this child is started
```
arielshaqed: What does this (empty) handler do?

local-proxy/src/index.ts
```diff
@@ -31,28 +23,24 @@ export async function startProxy(rootDir: string, localToken: string) {
...
   }).on('message', (message) => {
-    if (message === 'ready' && handle !== undefined) {
-      handle.send('server', server);
+    if (message === 'ready') {
+      // this child is ready
```
arielshaqed: This handler also seems empty.

local-proxy/src/server.ts
```diff
@@ -61,11 +60,6 @@ process.once('SIGUSR2', () => {
   process.kill(process.pid, 'SIGUSR2');
 });
 
-process.on('message', (m, netServer) => {
-  if (m === 'server') {
-    const httpServer = http.createServer(app);
-    (netServer as net.Server).on('connection', (socket) => httpServer.emit('connection', socket));
-  }
-});
+http.createServer(app).listen(19291, '127.0.0.1');
```
arielshaqed:
```suggestion
http.createServer(app).listen(port, '127.0.0.1');
```

local-proxy/package.json
```diff
@@ -7,9 +7,9 @@
   "description": "",
   "main": "dist/index.js",
   "scripts": {
-    "test": "echo \"No test specified\"",
+    "build": "rm -rf dist/ && tsc",
     "lint": "tslint -c ../common/tslint.yml -p .",
-    "build": "rm -rf dist/ && tsc"
+    "test": "ava -v dist/test/test_*.js"
```
arielshaqed: Nit: Can we just `ava dist/test/*.js`?  snake_caseIsForPython, I thought.

local-proxy/src/index.ts
```diff
@@ -1,22 +1,25 @@
...
+const isTestEnv = process.env.NODE_ENV === 'test';
 
-  await new Promise((resolve, reject) => {
-    server.listen(0, '127.0.0.1', resolve);
-    server.once('error',  reject);
-    return server;
-  });
-  const { port } = server.address() as net.AddressInfo;
+function log(message?: any, ...optionalParams: any[]) {
+  if (!isTestEnv) {
```
arielshaqed: Don't understand: log _except_ during tests?

local-proxy/src/index.ts
```diff
@@ -1,22 +1,25 @@
...
+    console.log(message, ...optionalParams);
+  }
+}
 
-  console.log(`Dev server listening on port: ${port}`);
+let port: number | undefined;
+const readyEmitter = new EventEmitter().setMaxListeners(1000);
+export function startProxy(rootDir: string, localToken: string) {
+  if (process.env.NODE_ENV !== 'test') {
```
arielshaqed: Double `if` on this `log`.

local-proxy/src/index.ts
```diff
@@ -29,33 +32,48 @@ export async function startProxy(rootDir: string, localToken: string) {
       SHIFT_DEV_SERVER_BASE_REQUIRE_PATH: path.resolve(path.join(rootDir, 'backend')),
       SHIFT_DEV_SERVER_LOCAL_TOKEN: localToken,
     },
+    // Workaround for tests:
+    // server.js uses babel/dir which has a console.log
+    // AVA transaltes console.log to stderr
```
arielshaqed: Nit:
```suggestion
    // AVA translates console.log to stderr
```

local-proxy/src/index.ts
```diff
@@ -29,33 +32,48 @@ export async function startProxy(rootDir: string, localToken: string) {
...
-    handle = child;
+  }).on('start', (_child) => {
+    if (process.env.NODE_ENV !== 'test') {
+      log('Local dev server started');
+    }
   }).on('message', (message) => {
-    if (message === 'ready' && handle !== undefined) {
-      handle.send('server', server);
+    if (message.type === 'ready') {
+      port = message.port;
```
arielshaqed: Might be interesting to log this.

local-proxy/src/index.ts
```diff
@@ -29,33 +32,48 @@ export async function startProxy(rootDir: string, localToken: string) {
...
+  const httpProxy = new proxy();
+  httpProxy.on('error', (err: any) => console.error(err.stack));
+  return (app: Application) => {
+    startProxy(rootDir, localToken);
+    app.use(async (req, res, next) => {
+      if (req.url !== '/invoke') {
+        return next();
+      }
+      if (!port) {
+        await new Promise((resolve) => readyEmitter.once('ready', resolve));
```
arielshaqed: Given that it's `once`, it might be easier to read if you defined the promise once and just `await`ed that same promise each time here.

arielshaqed: I don't understand.  The _first_ time we set up a new server, we wait for it to emit a `ready` event.  On every subsequent request, we _don't_ -- going to the wrong port until it is up.
So when the proxy is hit between when nodemon receives `restart` and `start`, it uses `port === undefined`.  The proxy target is the string `'localhost:undefined'`, and nastiness ensues.  In particular, if the subprocess fails to start up (e.g. a `require` or something else in global context before sending `start` kills it), I think the proxy will fail in an unpleasant way.

local-proxy/src/test/test_flakyInvoke.ts
```diff
@@ -0,0 +1,32 @@
...
+  const { port } = server.address() as AddressInfo;
+  const requests = [];
+  for (let i = 0; i < 100; i++) {
+    const req = got.post(`http://127.0.0.1:${port}/invoke`, {
+      body: { path: 'dummyBackend.js', args: [], handler: 'hello', },
+      json: true,
+    });
+    requests.push(req);
+  }
+  const responses = await Promise.all(requests);
```
arielshaqed: Consider
```ts
const responses = await Promise.all(
  Array.fill(100)
    .map(() => got.post(...))
);
```

I think it's easier to read because it has fewer mutations.

local-proxy/src/test/test_flakyInvoke.ts
```diff
@@ -0,0 +1,32 @@
...
+import { AddressInfo } from 'net';
+import express from 'express';
+import { setupProxy } from '../index';
+import test from 'ava';
+import path from 'path';
+import got from 'got';
+
+// This test should be considered NOT flaky
+// This means that if a test run fails there is most likely a bug in the code
+test('Flaky http proxy', async (t) => {
```
arielshaqed: What does this test?  AFAICT the proxy itself _isn't_ flaky here; could you retitle or comment?

local-proxy/package.json
vogre: Normalized as previously discussed in office, tests live in `src/test` and their format must be `/.*\.(test|spec)\.ts$/`.

local-proxy/src/index.ts
vogre: Yep, tests should only output success or failure (in verbose mode also the list of tests). The justification is in https://github.com/binaris/shiftjs/blob/ee6e17d3f3ec371ebd9aa0ad2182160cc103bdc2/common/config/rush/command-line.json#L83.

arielshaqed: I think I can confirm what you're saying below.

If I'm reading correctly (ironically, this entire text about inadvertent output writing) gets highlighted like an error in GitHub...) this applies only to standard output.  Tests are still allowed to write to standard output, but then Ava moves standard output to standard error.  I think.

local-proxy/src/index.ts
```diff
@@ -29,33 +41,47 @@ export async function startProxy(rootDir: string, localToken: string) {
...
-
   nodemon.on('quit', () => {
     process.exit();
-  }).on('start', (child) => {
-    handle = child;
-  }).on('message', (message) => {
-    if (message === 'ready' && handle !== undefined) {
-      handle.send('server', server);
+  }).on('start', (_child: any) => {
+    log('Local dev server started');
```
arielshaqed: This should go to standard error.

local-proxy/src/index.ts
```diff
@@ -29,33 +41,47 @@ export async function startProxy(rootDir: string, localToken: string) {
...
+  }).on('message', (message: any) => {
+    if (message.type === 'ready') {
+      portEmitter.emit('port', message.port);
     }
-  }).on('restart', (files) => {
-    console.log('Local dev server restarted due to changes in: ', files);
+  }).on('crash', () => {
+    log('Local dev server crashed');
```
arielshaqed: This should _definitely_ go to standard error.

PR #60, opened 2019-07-01
 Add rush test command

vogre:
Changes to #51 - replace bash with a rush command.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,12 +27,7 @@ jobs:
       #     command: npm run lint
       - run:
           name: Run tests
-          command: |
-            for f in $(find . -maxdepth 2 -name package.json); do
-              project=$(dirname $f)
-              echo Testing $project
-              (cd $project && npm test) || exit 1
-            done
+          command: node common/scripts/install-run-rush.js test --verbose
 workflows:
   version: 2
   build_and_test:
--- a/babel-macro/package.json
+++ b/babel-macro/package.json
@@ -22,6 +22,7 @@
     "@types/node": "^10.14.9",
     "babel-plugin-tester": "^6.4.0",
     "jest": "^24.8.0",
+    "jest-standard-reporter": "^1.0.1",
     "typescript": "^3.5.2"
   }
 }
--- a/babel-macro/package.json
+++ b/babel-macro/package.json
@@ -7,7 +7,7 @@
   ],
   "description": "",
   "scripts": {
-    "build": "rm -rf dist && tsc && cp -r src/test/backend/ dist/test/",
+    "build": "rm -rf dist && tsc && cp -r src/test/backend dist/test/",
     "test": "jest"
   },
   "author": "",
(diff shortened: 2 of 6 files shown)


.circleci/config.yml
```diff
@@ -27,12 +27,7 @@ jobs:
       #     command: npm run lint
       - run:
           name: Run tests
-          command: |
-            for f in $(find . -maxdepth 2 -name package.json); do
-              project=$(dirname $f)
-              echo Testing $project
-              (cd $project && npm test) || exit 1
-            done
+          command: node common/scripts/install-run-rush.js test --verbose
```
arielshaqed: This is much nicer.  Might even work with Windows!

babel-macro/package.json
```diff
@@ -7,7 +7,7 @@
   ],
   "description": "",
   "scripts": {
-    "build": "rm -rf dist && tsc && cp -r src/test/backend/ dist/test/",
+    "build": "rm -rf dist && tsc && cp -r src/test/backend dist/test/",
```
arielshaqed: Don't modern `tsc`s clean up enough to do this nicely and quickly?

common/changes/@binaris/shift-babel-macro/test-command_2019-07-01-09-13.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "comment": "",
```
arielshaqed: ?

common/config/rush/command-line.json
```diff
@@ -122,6 +122,12 @@
     //    */
     //   "shellCommand": "node common/scripts/my-global-command.js"
     // }
+    {
+        "commandKind": "bulk",
+        "name": "test",
+        "summary": "Run all tests",
+        "enableParallelism": false
```
arielshaqed: Why not?  (It would be disappointing if Rush is incapable of re-ordering the output to get this to print errors nicely.)

arikmaor: Can this be true?

babel-macro/package.json
arikmaor: why did you make this change? it can cause a `dist/test/backend/backend` to be created

common/changes/@binaris/shift-babel-macro/test-command_2019-07-01-09-13.json
arikmaor: Standard `rush` stuff...

common/config/rush/command-line.json
vogre: It is disappointing, really. I documented this and other reasons why it's serial now.

babel-macro/package.json
arielshaqed: But the behaviour this implements _is_ confusing Seems like we're trying to fix Microsoft/TypeScript#30602 _en passant_.  It will confuse various tools (such as anything that looks at file `mtime`s.  It discourages `tsc -w`.

I don't want to second-guess TypeScript.

vogre: It is not an uncommon pattern (e.g. https://github.com/ReactiveX/rxjs-tslint/blob/11d64176b717c5029c212d883bfc2dfc5d14d06d/package.json#L14), I just copied it from somewhere - this is just the most paranoid version of starting from a clean slate. I run `tsc -w` without cleaning directory during development but I agree it's not the best solution.

PR #61, opened 2019-07-01
Add linting

michaeladda:


local-proxy/package.json
```diff
@@ -29,6 +30,7 @@
     "@types/nanoid": "^2.0.0",
     "@types/node": "^10.14.9",
     "@types/rimraf": "^2.0.2",
+    "tslint": "~5.18.0",
```
vogre: Note this rush peculiarity, it prefers tilde to caret, can you change to `rush add --caret --dev -p tslint`? Although, this will probably become irrelevant when we change to automatic updates.

fetch-runtime/package.json
```diff
@@ -5,8 +5,14 @@
   "main": "index.js",
   "scripts": {
     "test": "echo \"No test specified\"",
+    "lint": "eslint -c ../common/eslintrc.js .",
```
vogre: I think we should convert `fetch-runtime` to typescript at some point because the `eslint` is similar, but different in places from `tslint`.

PR #62, opened 2019-07-01
Use typescript in fetch runtime

michaeladda:
tested manually by importing a newly generated packed tgz

### DIFF
--- a/fetch-runtime/package.json
+++ b/fetch-runtime/package.json
@@ -2,17 +2,16 @@
   "name": "@binaris/shift-fetch-runtime",
   "version": "0.0.1",
   "description": "",
-  "main": "index.js",
+  "main": "dist/index.js",
   "scripts": {
     "test": "echo \"No test specified\"",
-    "lint": "eslint -c ../common/eslintrc.js .",
-    "build": "echo \"nothing to do\""
+    "lint": "tslint -c ../common/tslint.yml -p .",
+    "build": "rm -rf dist/ && tsc"
   },
   "author": "",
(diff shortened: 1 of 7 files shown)


fetch-runtime/package.json
```diff
@@ -2,17 +2,16 @@
...
   "version": "0.0.1",
   "description": "",
-  "main": "index.js",
+  "main": "dist/index.js",
   "scripts": {
     "test": "echo \"No test specified\"",
-    "lint": "eslint -c ../common/eslintrc.js .",
-    "build": "echo \"nothing to do\""
+    "lint": "tslint -c ../common/tslint.yml -p .",
+    "build": "rm -rf dist/ && tsc"
```
arielshaqed: I'm discouraged that we repeatedly work against TypeScript; see https://github.com/binaris/shiftjs/pull/60/files#r299047990.

But it's not for this PR, given that we consistently use it throughout.

vogre: Do you think it's better to add a `clean` script, and then run `rushx clean && rushx build`?

arielshaqed: It's much more `make`y, and I think I prefer it.

In fact, given that you "_shouldn't be able to_" use artefacts from previous TypeScript runs (except via `npm link` or incorrectly distributing from there or misconfiguring your IDE or ...), in theory running it should never be necessary.  (And the difference between theory and practice is that there is none, in theory.)

michaeladda:
Thanks reviewers. Will address all requests in follow-up PRs.

PR #64, opened 2019-07-02
Add bulk clean

michaeladda:
As requested in #62

also alphabetized the scripts. sorry :-)

### DIFF
--- a/common/changes/@binaris/shift-babel-macro/master_2019-07-02-12-37.json
+++ b/common/changes/@binaris/shift-babel-macro/master_2019-07-02-12-37.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@binaris/shift-babel-macro",
+      "comment": "Add clean command",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@binaris/shift-babel-macro",
+  "email": "michael@binaris.com"
+}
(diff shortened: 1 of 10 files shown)


common/changes/@binaris/shift-babel-macro/master_2019-07-02-12-37.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@binaris/shift-babel-macro",
+      "comment": "Add clean command",
+      "type": "patch"
```
vogre: I think this can be a `"type": "none"` change (`rush change` and Enter without describing changes) since it doesn't affect outside users, and relates only to the build process and won't add noise to the changelog.

vogre:
Part of the issue is `rush` not yet having support for `watch` (https://github.com/microsoft/web-build-tools/issues/1202), one of the comments mentions that the standard `build` is a `gulp --clean`. In any case we can add a `clean` target and `build:watch` targets, but I think we should keep `build` as-is.

michaeladda:
Unfortunately, reading https://rushjs.io/pages/commands/rush_rebuild/ and https://rushjs.io/pages/commands/rush_build/ indeed, adding an explicit clean indeed seems counter productive. Closing. @arielshaqed let's discuss this off line if the current state is unsatisfactory.

arielshaqed:
Yup, JavaScript tooling is broken on this. It seems those who do not know Lisp are doomed to reimplement [asdf](https://common-lisp.net/project/asdf/asdf.html), poorly.

PR #66, opened 2019-07-03
Auto-update npm dependencies (with PR)

arikmaor:
This PR included created a user in github for circle-ci integration (@shiftjs-circleci)
This github user has a `user ssh key` and an `apikey` add to `circle-ci`
The `user ssh key` is used to `push` the updated `package.json` and `shrinkwrap.yaml` files.
The `apikey` is used to create a PR with `github` rest api.
To view the credentials: `aws s3 cp s3://secrets.binaris/github-shift-circleci - | less`


.circleci/config.yml
```diff
@@ -28,8 +28,34 @@ jobs:
...
           name: Run tests
           command: node common/scripts/install-run-rush.js test --verbose
+  update_npm_dependencies:
+    docker:
+      - image: circleci/node:10
+    steps:
+      - checkout
+      - run:
+          name: Install needed npm packages
+          command: npm install npm-check-updates @octokit/rest --no-save
```
vogre: While this is not recommended by rush (it introduces [phantom dependencies](https://rushjs.io/pages/advanced/phantom_deps/)), it should be safe enough in this flow.

.circleci/config.yml
```diff
@@ -28,8 +28,34 @@ jobs:
...
+      - run:
+          name: Configure git
+          command: |
+            git config user.email "auto-npm-dep-update@shiftjs.com"
```
vogre: Is this email activated, or supposed to be a noreply one?

common/scripts/update-npm-dependencies.js
```diff
@@ -0,0 +1,141 @@
...
+  updateRushShrinkwrapFile();
+  await pushChanges();
+  await createPullRequest();
+
+  console.info('Done!');
+}
+
+run().catch((error) => {
+  console.error(error);
+  process.exit(-1);
```
vogre: -1 is not a really good exit code, since it becomes 255, better `process.exit(1)`

common/scripts/update-npm-dependencies.js
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  updateRushShrinkwrapFile();
+  await pushChanges();
+  await createPullRequest();
+
+  console.info('Done!');
+}
+
+run().catch((error) => {
```
vogre: Can you add a `process.exit(0)` on success, to avoid issues with process not exiting if any timers/sockets are leftover?

common/scripts/update-npm-dependencies.js
```diff
@@ -0,0 +1,141 @@
+const ncu = require('npm-check-updates');
+const Octokit = require('@octokit/rest');
+const { installAndRun, findRushJsonFolder, RUSH_JSON_FILENAME } = require("./install-run");
+const fs = require('fs');
+const { join } = require('path');
+const { promisify } = require('util');
+const { spawnSync } = require('child_process');
+const readdir = promisify(fs.readdir);
```
vogre: Can you use `readdirSync` instead? This code is intended to be run as a script and the sync versions are really best suited for scripts.

common/scripts/update-npm-dependencies.js
```diff
@@ -0,0 +1,141 @@
...
+const exists = promisify(fs.exists);
```
vogre: Similarly, can you use `existsSync`? `exists` is deprecated due to not having error first callback, and the promisify only works due to `util.promisify.custom` which changes the behavior.

common/scripts/update-npm-dependencies.js
```diff
@@ -0,0 +1,141 @@
+const ncu = require('npm-check-updates');
+const Octokit = require('@octokit/rest');
+const { installAndRun, findRushJsonFolder, RUSH_JSON_FILENAME } = require("./install-run");
```
vogre: The double quotes hint at a lack of lint. Perhaps it would be worth it to move the script into a project and have the common `lint` framework, also benefit from having it's dependencies.

.circleci/config.yml
arikmaor: no-reply

common/scripts/update-npm-dependencies.js
arikmaor: done

common/scripts/update-npm-dependencies.js
arikmaor: We shouldn't have this issues.
If you do get them, I would like to notice them

common/scripts/update-npm-dependencies.js
arikmaor: changed

common/scripts/update-npm-dependencies.js
arikmaor: changed

.circleci/config.yml
arikmaor: I've had the same though while writing this

common/scripts/update-npm-dependencies.js
arikmaor: new project in this repo or in a new repo?

common/scripts/update-npm-dependencies.js
```diff
@@ -0,0 +1,126 @@
...
+    stdio: 'inherit'
+  });
+
+  if (status !== 0) {
+    throw new Error(`'${commandString}' exited with code ${status}`);
+  }
+}
+
+async function updatePackageFiles() {
+  const allPackageFiles = readdirSync(process.cwd(), { withFileTypes: true })
```
vogre: Come to think of it we should be able to get project directories out of rush.json - since rush repos often use depth of two for projects then this can be a general package that works with any rush repo.

arikmaor: yes, this is what i meant by "separate" repo in a previous comment.
but then the `tsconfig` will not be shared ;)

vogre: It can be in this repo and published to a package that is not shiftjs-...

arikmaor: Moved everything to a new repo/package, so we need to merge https://github.com/binaris/rush-update/pull/1 before merging this

does look nicer now

.circleci/config.yml
```diff
@@ -28,8 +28,31 @@ jobs:
...
+    steps:
+      - checkout
+      - run:
+          name: Configure git
+          command: |
+            git config user.email "auto-npm-dep-update@shiftjs.com"
+            git config user.name "autoupdate"
+      - run:
+          name: Update npm dependencies
+          command: node common/scripts/install-run.js rush-update@beta rush-update -x @types/node --repo-owner binaris --repo-name shiftjs
```
vogre: This is really cool that all the logic is inside one script, just remember to remove the `@beta`.

arikmaor: forgot to push :P

tools/rush-update/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "main": "dist/main.js",
+  "bin": {
+    "rush-update": "dist/cli.js"
+  },
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "tsc",
+    "lint": "tslint -p ."
```
vogre: Possible to reuse common tslint like here https://github.com/binaris/shiftjs/blob/ee6e17d3f3ec371ebd9aa0ad2182160cc103bdc2/db/package.json#L12.

arikmaor: Changed and also removed all obsolete `tslint.yml` files

PR #71, opened 2019-07-08
Run `rush update` after auto-updating dependencies

arikmaor:


tools/rush-update/src/main.ts
```diff
@@ -1,11 +1,12 @@
 import { updatePackageFiles } from './update-npm';
-import { getProjectFolders, updateRushShrinkwrapFile } from './rush';
-import { commitCurrentChangesToBranch, pushToRemoteBranch } from './git';
+import { getProjectFolders, updateRushShrinkwrapFile, generateRushChangeFiles } from './rush';
+import { switchToNewBranch, commitChanges, pushToRemoteBranch } from './git';
 import { createPullRequest } from './github';
 import { NCUParams } from 'npm-check-updates';
 
 export const DEFAULT_BRANCH = 'update-npm-dependencies';
```
vogre: Just noticed that the branch name is constant, do you think it would work better with a dynamic branch like `update-npm-dependencies-2019-07-08`, otherwise every PR needs to be reviewed or the following ones will fail, even during a weekend.

tools/rush-update/src/rush.ts
```diff
@@ -22,10 +23,16 @@ export const getProjectFolders = once(() => getRushConfig()
   .then(({projects}) => projects.map((p) => p.projectFolder))
 );
 
-export async function updateRushShrinkwrapFile() {
+export async function updateRushShrinkwrapFile(): Promise<void> {
```
vogre: Does this function have to be async?

tools/rush-update/src/rush.ts
```diff
@@ -22,10 +23,16 @@ export const getProjectFolders = once(() => getRushConfig()
...
   const { installAndRun } = await importInstallRunScript();
   const statusCode = installAndRun('@microsoft/rush', await getRushVersion(), 'rush', ['update', '--full']);
   if (statusCode !== 0) {
     throw new Error(`'rush update' exited with code ${statusCode}`);
   }
 }
+
+export function generateRushChangeFiles(): void {
+  execSync('yes "" | rush change', {
```
vogre: Can you add a comment that this answers the question `Describe changes, or ENTER if no changes:` with ENTER?

tools/rush-update/src/main.ts
arikmaor: maybe we should fine tune it a bit (to run only once on the weekend)
but anyway, I'm for it failing at some point if a PR/branch already exists

tools/rush-update/src/rush.ts
arikmaor: no `async` function has to be an async function, every `const x = await foo()` can be replaced with `foo().then(x => {...})`

this one `await`s 2 times

tools/rush-update/src/rush.ts
arikmaor: adding

PR #79, opened 2019-07-09
Implement query language

bergundy:


common/changes/@binaris/shift-db/query-language_2019-07-09-09-03.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@binaris/shift-db",
+      "comment": "Implement query language",
+      "type": "major"
```
vogre: This gives version 1.0.0 which opts-in into semver, until now we tried to only use patch versions, do you think we should switch?

db/src/test/db.match.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+});
+
+test('startsWith matches string', (t) => {
+  t.false(match({ key: 'abc', value: null }, Q.value.startsWith('ab')));
+  t.false(match({ key: 'abc', value: 0 }, Q.value.startsWith('ab')));
+  t.true(match({ key: 'abc', value: 'abc' }, Q.value.startsWith('ab')));
+  t.false(match({ key: 'abc', value: 'ABC' }, Q.value.startsWith('ab')));
+});
+
+test('and applies all filters', (t) => {
```
arielshaqed:
```suggestion
test('any applies all filters', (t) => {
```

db/src/test/db.match.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+  t.true(match({ key: 'abc', value: 0 }, Q.all(Q.key.startsWith('ab'), Q.value.eq(0))));
+  t.false(match({ key: 'abc', value: 0 }, Q.all(Q.key.startsWith('ab'), Q.value.eq(1))));
+  t.false(match({ key: 'abc', value: 0 }, Q.all(Q.key.eq('ab'), Q.value.eq(0))));
+  t.false(match({ key: 'abc', value: 0 }, Q.all(Q.key.eq('ab'), Q.value.eq(1))));
+});
+
+test('or applies all filters', (t) => {
```
arielshaqed:
```suggestion
test('any applies all filters', (t) => {
```

db/src/test/query.spec.ts
```diff
@@ -0,0 +1,296 @@
+import test from 'ava';
+import { IllegalArgumentError } from '../errors';
+import { key, all, any, not, value, typedValue, filter, ASC, DESC } from '../query';
+
+function stringifyParse(v: any) {
```
arielshaqed:
```suggestion
function parseStringify(v: any) {
```
(or `stringifyThenParse`, given that some people read it each way).

db/src/query.ts
```diff
@@ -0,0 +1,441 @@
...
+  }
+  for (const f of filters) {
+    if (!f[filterSymbol]) {
+      throw new TypeError('Given filter is invalid');
+    }
+  }
+}
+
+export function all(...filters: NonEmptyArray<Filter>): AndFilter {
+  checkFilters(...filters);
```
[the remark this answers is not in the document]
bergundy: After our F2F we agreed that it would be very hard for a user to pass a filter created not with one of the exposed functions.
In the future the filters will be serialized and sent over network which will make it impossible to verify filters were created using one of the provided constructors.

db/src/db.ts
```diff
@@ -91,4 +96,84 @@ export class DB {
...
+    const sortedResults = orderBy ? sortWith(orderBy.map(buildComparator), results) : results;
+    return sortedResults.slice(skip, limit === undefined ? undefined : (skip || 0) + limit);
+  }
+}
+
+export function buildComparator([p, direction]: Q.Order) {
+  return direction === Q.ASC ? ascend(path(p)) : descend(path(p));
+}
+
+export function match(doc: Document, filter: Q.Filter): boolean {
```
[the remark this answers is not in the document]
arielshaqed: Discussed F2F, OK now.

db/src/db.ts
```diff
@@ -91,4 +96,84 @@ export class DB {
...
+        return false;
+      }
+      const regexp = new RegExp(filter.pattern, filter.caseInsensitive ? 'i' : undefined);
+      return regexp.test(value);
+    case 'startsWith':
+      if (typeof value !== 'string') {
+        return false;
+      }
+      return value.startsWith(filter.value);
+  }
```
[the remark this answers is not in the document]
arielshaqed: Yes, but if the client passes a filter from a different version to the server it might happen.  (E.g. consider using a serialized filter that comes from the future, because it's from a newer package that uses a newer version.  Or _something_.)
Let's have an exception / error report when that happens, please.  (Could be at an earlier validation phase too, of course)

db/src/query.ts
```diff
@@ -0,0 +1,441 @@
...
+
+export type Filter = EqFilter | NeFilter
+  | GtFilter | GteFilter | LtFilter | LteFilter
+  | ExistsFilter | IsNullFilter
+  | MatchesFilter | StartsWithFilter
+  | AndFilter | OrFilter | NotFilter;
+
+const proxyHandler = {
+  get(obj: Path, prop: string) {
+    return prop in obj ? obj[prop as keyof Path] : obj.field(prop);
```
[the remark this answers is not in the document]
arielshaqed: True, sorry.  Document it explicitly somewhere?

db/src/query.ts
```diff
@@ -0,0 +1,444 @@
...
+      return {
+        [filterSymbol]: true,
+        operator: 'matches',
+        path: this.parts,
+        pattern,
+        caseInsensitive,
+      };
+    } else if (pattern instanceof RegExp) {
+      if (/[^i]/.test(pattern.flags)) {
+        throw new TypeError('Given RegExp flags not supported');
```
arielshaqed:
```suggestion
        throw new TypeError('Only /i RegExp flag supported');
```
I prefer to state what is allowed, for about the same character count.  Because even an error message can be positive.

db/src/query.ts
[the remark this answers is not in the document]
arielshaqed: Agreed.  It is not possible to write (reasonable) code using the API that allows an injection using punning (passing an object where a string or number was expected), as values only go in leaves.

db/src/db.ts
```diff
@@ -135,14 +135,22 @@ export function buildComparator([p, direction]: Q.Order) {
   return direction === Q.ASC ? ascend(path(p)) : descend(path(p));
 }
 
+export function wrappedMatch(doc: Document, filter: Q.Filter): boolean {
```
arielshaqed: Cool.  Please add a comment why we need this (so I don't delete it as "unnecessary cruft").

PR #81, opened 2019-07-09
Disable auto updates of package.json dependencies

bergundy:


tools/rush-update/src/main.ts
```diff
@@ -45,10 +47,9 @@ export default async function main({
   prBody?: string,
 }) {
   const projectFolders = await getProjectFolders();
-  const shouldUpdateShrinkwrapFile = await updatePackageFiles(projectFolders, ncuParams || {});
-  if (!shouldUpdateShrinkwrapFile) {
+  if (checkUpdates) {
+    await updatePackageFiles(projectFolders, ncuParams || {});
     console.info('All dependencies are up to date!');
-    return;
```
vogre: This breaks in the case when there are no updated dependencies, also the new behavior doesn't check whether `rush update --full` generated any changes and tries to create an empty commit / PR...

arikmaor:
After F2F in the room, we've decided at this point to only make minor changes instead of trying to optimise the process completely #85

PR #85, opened 2019-07-10
Run update_npm_dependencies weekly instead of nightly

arikmaor:
need to run `rush publish -a -b master -p --npm-auth-token $(aws s3 cp s3://secrets.binaris/npm -)` after merging

### DIFF
--- a/common/changes/rush-update/change-auto-update-dep-job_2019-07-10-09-22.json
+++ b/common/changes/rush-update/change-auto-update-dep-job_2019-07-10-09-22.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "rush-update",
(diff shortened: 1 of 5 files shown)


common/changes/rush-update/change-auto-update-dep-job_2019-07-10-09-22.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "rush-update",
+      "comment": "Request pr review from michael when auto-updating dependencies",
```
vogre: I think the change to the package of rush-update is "Introduced new flag --pr-reviewers", the reviewer in this case is part of configuration and not the change.

PR #89, opened 2019-07-10
Change db.update API and Serializable type

bergundy:


db/src/index.ts
```diff
@@ -42,10 +42,10 @@ export async function remove(key: string): Promise<boolean> {
  * @param initializer - `updater` will get this value if no document exists for `key`.
  * @return - The new value returned from updater
  */
-export async function update<T extends Serializable, R extends Serializable>(
-  key: string, updater: (state?: T) => R, initializer?: T
+export async function update<R extends Serializable>(
+  key: string, updater: (state?: Serializable) => R
 ): Promise<R> {
-  return await db.update(key, updater, initializer);
+  return await db.update(key, updater);
```
arielshaqed: Why is initializer gone?

db/src/index.ts
```diff
@@ -42,10 +42,10 @@ export async function remove(key: string): Promise<boolean> {
  * @param initializer - `updater` will get this value if no document exists for `key`.
```
arielshaqed: Why is initializer still here?

db/src/index.ts
arikmaor: there is no need for it
it is the same as using ES6 default value

PR #91, opened 2019-07-11
Unify tsconfigs by extending a common config

vogre:
Preparation before adding yet another tsconfig.


vogre:
>  Do you think we should specify include only src?

I think we should try having [project references](https://www.typescriptlang.org/docs/handbook/project-references.html) to separate `src/` and `test/`, but until we have something like an `examples/` directory which we don't want to compile it's of no use.

PR #93, opened 2019-07-14
Implement subscriptions in DB module

bergundy:


db/src/db.ts
```diff
@@ -85,17 +147,55 @@ export class DB {
...
+  public async poll(keysToVersions: KeyedVersions, opts: Partial<PollOptions> = {}): Promise<KeyedPatches> {
+    const patchPromise = new Promise<KeyedPatches>((resolve) =>
+      this.once('patch', (key, patch) => resolve([[key, [patch]]])));
+    const keyedPatchesOrUndef = await Promise.all(keysToVersions.map(async ([key, version]) => {
+      const doc = await this.getWithMeta(key);
+      if (doc === undefined) {
+        return undefined;
+      }
+      const patches = doc.patches.filter((patch) => patch.version > version);
+      if (patches.length === 0) {
```
arielshaqed: How can this happen?  (If it cannot, should we report or monitor if it does?)

db/src/db.ts
```diff
@@ -85,17 +147,55 @@ export class DB {
...
 
+  public async poll(keysToVersions: KeyedVersions, opts: Partial<PollOptions> = {}): Promise<KeyedPatches> {
+    const patchPromise = new Promise<KeyedPatches>((resolve) =>
+      this.once('patch', (key, patch) => resolve([[key, [patch]]])));
+    const keyedPatchesOrUndef = await Promise.all(keysToVersions.map(async ([key, version]) => {
+      const doc = await this.getWithMeta(key);
+      if (doc === undefined) {
+        return undefined;
+      }
+      const patches = doc.patches.filter((patch) => patch.version > version);
```
arielshaqed: If `patch.version`s are increasing, this might be (much) more efficient as `find` and then `slice`.

db/src/db.ts
```diff
@@ -85,17 +147,55 @@ export class DB {
...
+          updatedAt: Date.now(),
+        }));
+        this.emit('patch', key, patch);
+      }
+      return nextValue;
     });
   }
 
+  public async poll(keysToVersions: KeyedVersions, opts: Partial<PollOptions> = {}): Promise<KeyedPatches> {
+    const patchPromise = new Promise<KeyedPatches>((resolve) =>
```
arielshaqed: I don't understand how this return value satisfies either part (keys or versions of those keys) of `keysToVersions`.

db/src/test/db.spec.ts
```diff
@@ -31,11 +32,13 @@ test('DB.get returns undefined when no key exists', async (t) => {
...
 
-test('DB.create creates a new document and returns true', async (t) => {
+test('DB.create creates a new document, returns true, and sets version to 1', async (t) => {
   const { db } = t.context;
+  const t0 = Date.now();
   const ret = await db.create('test', { a: 1 });
-  const val = await db.get('test');
-  t.deepEqual(val, { a: 1 });
+  const doc = await db.getWithMeta('test');
+  t.deepEqual(omit(['updatedAt'], doc), { value: { a: 1 }, version: 1, patches: [] });
```
arielshaqed: If we injected a `nowFunc` into `db` we could verify it calls `nowFunc` correctly (e.g. it could increase its return value by 13 on each call).

db/src/test/db.spec.ts
```diff
@@ -73,24 +76,161 @@ test('DB.remove removes existing key from DB and returns true', async (t) => {
...
+    ]],
+  ]);
+});
+
+test('DB.create works after remove', async (t) => {
+  const { db } = t.context;
+  const t0 = Date.now();
+  await db.create('test', 7);
+  await db.remove('test');
+  t.true(await db.create('test', 8));
```
arielshaqed: This test may fail on some eventually consistent datastores: it's perfectly reasonable for the second `create` to fail because the key still exists.

db/src/utils.ts
```diff
@@ -0,0 +1,20 @@
+export class TimeoutError extends Error {
+  public readonly name = 'TimeoutError';
+}
+
+// Tests not included since this code is borrowed from a utility library
```
arielshaqed: Y U no publish library?! ლ(ಠ益ಠლ)

common/changes/@binaris/shift-db/subscriptions-db_2019-07-14-13-56.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@binaris/shift-db",
+      "comment": "Implement subscription functions",
+      "type": "patch"
```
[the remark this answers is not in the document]
vogre: A major version would push us to 1.0.0, so we should use patch versions until we are ready (see comments on #79).

db/src/test/db.spec.ts
bergundy: I understand but this is good enough IMHO, it gets more complex when it's used for versions (see new changes).

db/src/test/db.spec.ts
bergundy: Not on the local DB though..

db/src/utils.ts
bergundy: Not worth it ATM

db/src/test/db.spec.ts
arielshaqed: OK.

db/src/db.ts
```diff
@@ -1,39 +1,137 @@
...
+  return [x, y + amount];
+}
+
+export function decrVersion([x, y]: Version, amount: number = 1): Version {
+  return [x, y - amount];
+}
+
+export interface Versioned<T> {
+  /**
+   * A document starts with version 1, each time a document is updated the version is incremented.
```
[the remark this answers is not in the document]
arielshaqed: Also maybe export a versionCompare, it will be useful for clients I think.

db/src/index.ts
```diff
@@ -42,9 +42,9 @@ export async function remove(key: string): Promise<boolean> {
  * @return - The new value returned from updater
  */
 export async function update<T extends Serializable = any>(
-  key: string, updater: (state?: T) => T
+  key: string, updater: (state?: DeepReadonly<T>) => T, options?: UpdateOptions,
 ): Promise<T> {
```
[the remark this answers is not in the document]
bergundy: See changes after our F2F.

db/src/db.ts
```diff
@@ -1,39 +1,137 @@
...
+   */
+  readonly version: Version;
+  readonly value: T;
+}
+
+export interface Patch {
+  readonly version: Version;
+  /**
+   * Used to mark an update originated from a specific operation (useful for optimistic UI).
```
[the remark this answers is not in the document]
arielshaqed: OK, so recommended to be a client-chosen UUID.  That works.

db/src/db.ts
```diff
@@ -1,39 +1,137 @@
...
+    const ops = compare({ root: prevValue }, { root: value });
+    if (ops.length > 0) {
+      const patch = { version: nextVersion, ops, ...options };
+      await this.db.put(key, JSON.stringify({
+        version: nextVersion,
+        value,
+        patches: patches.slice(-NUM_PATCHES_TO_KEEP).concat(patch),
+        updatedAt: hrnano(),
+      }));
+      this.emit('patch', key, patch);
```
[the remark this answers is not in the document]
arielshaqed: OK as this is used only by the poller -- which doesn't care!

PR #96, opened 2019-07-17
Expose subscription functions via local proxy

bergundy:


common/changes/@binaris/shift-db/subscriptions-local-proxy_2019-07-17-12-50.json
```diff
@@ -0,0 +1,16 @@
...
+      "type": "patch"
+    },
+    {
+      "packageName": "@binaris/shift-db",
+      "comment": "Export getVersioned() and poll()",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@binaris/shift-db",
+  "email": "roey.berman@gmail.com"
```
arielshaqed: Please keep 3133+ haxx0r5 away from our code, prefer `@shiftjs.com` :-)

common/changes/@binaris/shift-local-proxy/subscriptions-local-proxy_2019-07-17-12-50.json
```diff
@@ -0,0 +1,16 @@
...
+      "type": "patch"
+    },
+    {
+      "packageName": "@binaris/shift-local-proxy",
+      "comment": "Disallow requiring files outside of root dir",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@binaris/shift-local-proxy",
+  "email": "roey.berman@gmail.com"
```
arielshaqed: Again.

local-proxy/src/server.ts
```diff
@@ -1,4 +1,4 @@
-import { join as pathJoin } from 'path';
+import { join as pathJoin, resolve as pathResolve } from 'path';
```
arielshaqed: `import * as path from 'path'` would let you use `path.join` and `path.resolve`, with *_less typing_*!!1!

local-proxy/src/server.ts
```diff
@@ -66,10 +66,20 @@ app.post('/invoke', json(), async (req, res) => {
   const { path, handler, args } = req.body;
   registry.register(requestId);
   try {
-    await transpilePromise;
-    // Make sure we use tmpDir as absolute path
-    const mod = require(pathJoin(tmpDir, path));
-    const fn = mod[handler];
+    let fn: (...args: any[]) => any;
+    if (path === '@binaris/shift-db' && (handler === 'getVersioned' || handler === 'poll')) {
```
arielshaqed: This is not nice hard-coding.

```suggestion
    if (path === '@binaris/shift-db') {
      const shiftDb = require(path);
      ret = shiftDb[handler];
      return ret[magicExportToBackendSymbol] && ret;
    }
```

local-proxy/src/server.ts
```diff
@@ -66,10 +66,20 @@ app.post('/invoke', json(), async (req, res) => {
...
+      fn = require(path)[handler];
+    } else {
+      await transpilePromise;
+      const joinedDir = pathResolve(tmpDir, path);
+      if (!joinedDir.startsWith(tmpDir)) {
+        return res.status(403).send({
+          error: 'Tried to reference path outside of root dir',
```
arielshaqed:
```suggestion
          error: 'Cannot reference path outside of root dir',
```

Should we also say _which_ path it was, or will the backtrace be enough to show which file required which path?

local-proxy/src/server.ts
```diff
@@ -66,10 +66,20 @@ app.post('/invoke', json(), async (req, res) => {
   const { path, handler, args } = req.body;
   registry.register(requestId);
   try {
-    await transpilePromise;
-    // Make sure we use tmpDir as absolute path
-    const mod = require(pathJoin(tmpDir, path));
-    const fn = mod[handler];
+    let fn: (...args: any[]) => any;
+    if (path === '@binaris/shift-db' && (handler === 'getVersioned' || handler === 'poll')) {
+      fn = require(path)[handler];
```
vogre: This guarantees that `local-proxy` depends on `db`, why not put it in dependencies and import it on start?

local-proxy/src/server.ts
```diff
@@ -66,10 +66,20 @@ app.post('/invoke', json(), async (req, res) => {
...
+    } else {
+      await transpilePromise;
+      const joinedDir = pathResolve(tmpDir, path);
+      if (!joinedDir.startsWith(tmpDir)) {
```
vogre: Supposedly this check should be more robust for win32: https://github.com/binaris/shiftjs/blob/b1960e15c96075e0dcdce0b8c4797a6d9920673c/babel-macro/src/macro.ts#L105.

local-proxy/src/server.ts
bergundy: There's a variable called path in the code below, it would shadow the path module.

local-proxy/src/server.ts
bergundy: Sounds good

local-proxy/src/server.ts
bergundy: Going to do what @vogre suggested below.

local-proxy/src/server.ts
bergundy: That doesn't seem right..
What about `a/../../b`?

local-proxy/src/server.ts
bergundy: I'll add it.

local-proxy/src/server.ts
vogre: relative is the result of `path.relative` so it shouldn't contain `..`. This is taken from [SO](https://stackoverflow.com/questions/37521893/determine-if-a-path-is-subdirectory-of-another-in-node-js), maybe node `path-is-inside` is better, although it has quite a different implementation.

local-proxy/src/server.ts
```diff
@@ -19,9 +19,25 @@ const localToken = process.env.SHIFT_DEV_SERVER_LOCAL_TOKEN;
 if (!localToken) {
   throw new Error('SHIFT_DEV_SERVER_LOCAL_TOKEN env var not defined');
 }
+
+const whitelistedModulesArr = (process.env.SHIFT_DEV_SERVER_MODULES_WHITELIST || '@binaris/shift-db').split(',');
```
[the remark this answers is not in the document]
vogre: I think I would have preferred it to be a real import and not a generic require of whitelist, but it's early to tell how it's going to be used.

db/src/index.ts
```diff
@@ -55,3 +65,26 @@ export async function update<T extends Serializable = any>(
...
+  const doc = await db.getWithMeta<T>(key);
+  if (doc === undefined || doc.value === undefined) {
+    return undefined;
+  }
+  return {
+    version: doc.version,
+    value: doc.value,
+  };
+}
+getVersioned.__shiftjs__ = 'exposed';
```
vogre: Note that format changed in https://github.com/binaris/shiftjs/pull/94/files#diff-ba07279b055fab1a8980733eebc00d7aR39.

PR #99, opened 2019-07-18
Cleanup temp directories when ctrl-c is used in terminal 

vogre:
To avoid confusing leftover .shift_local_proxy_ directories


arielshaqed:
Nice !! :-)

PR #101, opened 2019-07-21
Change subscriptions backend interface

bergundy:


db/src/index.ts
```diff
@@ -75,9 +75,9 @@ export async function poll(keysToVersions: KeyedVersions): Promise<KeyedPatches>
 poll.__shiftjs__ = { exposed: true };
 
 /**
- * Gets a single document by key including its version.
+ * Gets a initial document in an intent to for poll on it.
  */
-export async function getVersioned<T extends Serializable = any>(key: string): Promise<Versioned<T> | undefined> {
+export async function getForPoll<T extends Serializable = any>(key: string): Promise<Versioned<T> | undefined> {
```
arielshaqed: Thinking about it, on the backend it's going to be more than a `get` operation.  I'd prefer to rerename.
```suggestion
export async function startPolling<T extends Serializable = any>(key: string): Promise<Versioned<T> | undefined> {
```

bergundy:
@arielshaqed changed some more things, PTAL

PR #102, opened 2019-07-21
Implement subscriptions frontend library

bergundy:


subscriptions/src/rxutils.ts
```diff
@@ -0,0 +1,24 @@
...
+    return pipe(
+      scan(
+        ({ state }: StateAndOptionalOutput<T, O>, input: I): StateAndOptionalOutput<T, O> => fn(state, input),
+        { state: initialState, output: undefined }),
+      map(({ output }) => output!),
+    );
+  }
+
+export function takeUntilLast<T, U>(obs$: Observable<U>): MonoTypeOperatorFunction<T> {
+  return takeUntil(obs$.pipe(last(undefined, null)));
```
arielshaqed: Can we just
```ts
takeUntil(obs$.pipe(ignoreElements()))
```
?

subscriptions/src/subscriptions.ts
```diff
@@ -0,0 +1,214 @@
...
+}
+
+type VersionedStateGetter<T> = (key: string) => Promise<Versioned<T>>;
+
+export const REGISTER: 'register' = 'register';
+export const DEREGISTER: 'deregister' = 'deregister';
+const UPDATE: 'update' = 'update';
+
+interface RegisterPollerInput {
+  readonly action: 'register';
```
arielshaqed: `typeof REGISTER`?

subscriptions/src/subscriptions.ts
```diff
@@ -0,0 +1,214 @@
...
+}
+interface UpdatePollerInput {
+  readonly action: 'update';
+  readonly keysToVersions: Array<[string, Version]>;
+}
+type PollerInput = RegisterPollerInput | DeregisterPollerInput;
+
+export type NonEmptyArray<T> = [T, ...T[]];
+
+export function isNonEmptyArray<T>(arr: ReadonlyArray<T>): arr is NonEmptyArray<T> {
```
arielshaqed: This is `const`-unsafe (if you'll excuse my C++): it transforms nonempty `ReadonlyArray`s into `Array`s.  AFAICT the parameter has to be `T[]`, or you need to define
```ts
export type NonEmptyReadonlyArray<T> = NonEmptyArray<T> & ReadonlyArray<T>
```
and use that (untested).

subscriptions/src/subscriptions.ts
```diff
@@ -0,0 +1,214 @@
...
+  input: PollerInput | UpdatePollerInput,
+): im.Map<string, Version> {
+  switch (input.action) {
+    case REGISTER:
+      return keysToVersions.set(input.key, input.version);
+    case DEREGISTER:
+      return keysToVersions.delete(input.key);
+    case UPDATE:
+      // Only update existing keys in case key was updated after deregister
+      return keysToVersions.withMutations((m) =>
```
arielshaqed: This looks like a reduction on `input.keysToVersions`, why `useMutations`?  E.g. (untested):
```ts
return input.keysToVersions.reduce((k2v, [k, v]) => k2v.update(k, (prev) => prev === undefined ? undefined : v), keysToVersions);
```

subscriptions/src/subscriptions.ts
```diff
@@ -0,0 +1,214 @@
...
+    return merge(
+      sharedObs$,
+      subject.pipe(takeUntilLast(sharedObs$)),
+    ).pipe(
+      scan(pollReducer, im.Map()),
+      filter((keysToVersions) => keysToVersions.size > 0),
+      // TODO: backoff on error
+      concatMap(poller),
+      map((patchesPerKey) =>
+        patchesPerKey.filter((keyedPatch): keyedPatch is KeyedNonEmptyPatch => isNonEmptyArray(keyedPatch[1]))),
```
arielshaqed: I don't understand why look at the second element of `keyedPatch` to declare `keyedPatch` is nonempty.

subscriptions/src/test/subscriptions.spec.ts
```diff
@@ -0,0 +1,191 @@
+import test from 'ava';
+import * as im from 'immutable';
+import { stub } from 'sinon';
+import { concat, EMPTY, NEVER, of, Subject } from 'rxjs';
+import { delay, take, tap, toArray } from 'rxjs/operators';
+import { Patch, Version } from '@binaris/shift-interfaces/dist/subscriptions';
+import {
+  DEREGISTER,
```
arielshaqed: Nit: Please order these symbols somehow.

subscriptions/src/test/subscriptions.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+  PollerInputWithObserver,
+  pollReducer,
+  publish,
+  publishInputMapper,
+  REGISTER,
+  serverUpdates,
+} from '../subscriptions';
+
+// tslint:disable-next-line:variable-name
+const Version = (major: number, minor: number): Version => [major, minor];
```
arielshaqed: If this were a `function` you wouldn't need to `tslint:disable`.

subscriptions/src/test/subscriptions.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+test('poll stops polling when asked to deregister key', async (t) => {
+  const poller = mockPoller();
+  await of(
+    { action: REGISTER, key: 'test1', version: Version(1, 1) },
+    { action: REGISTER, key: 'test2', version: Version(1, 2) },
+    { action: DEREGISTER, key: 'test1' },
+  ).pipe(
+    poll(poller),
+  ).toPromise();
+  t.is(poller.args.length, 3);
```
arielshaqed: Why not just one single `t.deepEqual`?  That would give more context during error messages (I think).

subscriptions/src/test/subscriptions.spec.ts
bergundy:
```python
True
```
```js
true
```

subscriptions/src/test/subscriptions.spec.ts
bergundy: Because of the `toJS()` aspect of it..

arielshaqed:
```
t.deepEqual(
  poller.args.map(([version, _]) => version.toJS(),
  [
    { test1: Version(1, 1) },
    { test1: Version(1, 1), test2: Version(1, 2) },
    { test2: Version(1, 2) },
  ]);
```
?

Even gives a name to that elusive element 0.

subscriptions/src/test/subscriptions.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+  t.deepEqual(poller.args[0][0].toJS(), { test1: Version(1, 1) });
+  t.deepEqual(poller.args[1][0].toJS(), { test1: Version(1, 1), test2: Version(1, 2) });
+  t.deepEqual(poller.args[2][0].toJS(), { test2: Version(1, 2) });
+});
+
+test('poll updates version on poll result', async (t) => {
+  const poller = mockPoller().resolves([['test', [{ version: Version(1, 2), ops: [] }]]]);
```
arielshaqed: Not sure we want to commit to a version update on an idempotent write.  Maybe change the stored value too, to make that clear?

db/src/db.ts
```diff
@@ -254,7 +254,7 @@ export class DB extends EventEmitter {
           return it.end(resolve);
         }
         const { value } = JSON.parse(rawValue);
-        if (wrappedMatch({ key, value }, filter)) {
+        if (value !== undefined /* Not a tombstone */ && wrappedMatch({ key, value }, filter)) {
```
arielshaqed: This doesn't belong on this PR, I think.

bergundy: My bad, but if you reviewed it I'm merging.

PR #107, opened 2019-07-30
Enable more tslint strict rules

vogre:
Discovered when writing an async function and forgetting to await

Also add await-promise rule

### DIFF
--- a/common/tslint.yml
+++ b/common/tslint.yml
@@ -3,6 +3,15 @@ extends:
     - 'tslint:recommended'
 jsRules:
 rules:
+    no-floating-promises: true
+    strict-boolean-expressions:
+      - true
+      - allow-mix
+      - allow-undefined-union
+      - allow-null-union
+      - allow-string
+      - allow-number
+      - ignore-rhs
     quotemark:
       - true
       - single
--- a/common/tslint.yml
+++ b/common/tslint.yml
@@ -3,6 +3,7 @@ extends:
     - 'tslint:recommended'
 jsRules:
 rules:
+    await-promise: true
     no-floating-promises: true
     strict-boolean-expressions:
       - true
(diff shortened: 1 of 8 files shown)


common/tslint.yml
```diff
@@ -3,6 +3,16 @@ extends:
...
+    await-promise: true
+    no-floating-promises: true
+    strict-boolean-expressions:
+      - true
+      - allow-mix
+      - allow-undefined-union
+      - allow-null-union
+      - allow-string
+      - allow-number
+      - ignore-rhs
```
arielshaqed: Cool!  (Also good to see TSLint invented a new way to write a very limited set of generics, https://palantir.github.io/tslint/rules/strict-boolean-expressions/)

shift-react-app/src/shiftReactApp.ts
```diff
@@ -7,7 +7,7 @@ import {
   ignoreShift,
 } from './steps';
 
-const steps = [
+const steps: Array<(() => Promise<string>) | (() => string) | (() => void)> = [
```
arielshaqed: Why is this suddenly required?

tools/rush-update/src/types.d.ts
```diff
@@ -5,5 +5,5 @@ declare module 'npm-check-updates' {
     upgrade?: boolean,
   }
 
-  function run(params: NCUParams): Record<string, string[]>;
+  function run(params: NCUParams): Promise<Record<string, string[]>>;
```
arielshaqed: This was a bug in the old type definitions, right?

vogre: Yep! The implementation correctly awaited.

shift-react-app/src/shiftReactApp.ts
vogre: It looks like typescript inferred that the array is of functions of the form `() => void` based on the first function in the array, and allows assigning `() => string` to a `() => void`.

Issue #104, opened 2019-07-25
Install with yarn is not supported


michaeladda:
requested in #103

PR #111, opened 2019-08-01
Serving function

vogre:
This does not yet implement the expected spec (missing caching, various routing rules, most tests), relies on `bn` commands, and misses edge-cases handling

Note that `@binaris/shift-db` files are excluded, since the package relies on local LevelDB packages, the remote deployment is expected to replace it with a remote implementation.


server-function/shift_deploy_binaris.sh
```diff
@@ -0,0 +1,57 @@
...
+done
+cp -a `ls | grep -v node_modules | grep -v backend` .shift_deploy
+cat >.shift_deploy/binaris.yml <<EOF
+functions:
+  public_shift_builtin:
+    file: function.js
+    entrypoint: handler
+    runtime: node8
+EOF
+cat >.shift_deploy/function.js <<EOF
```
michaeladda: why embed it in the shell script?
wouldn't keeping it separate make development easier (editors) and allow unit testing?

server-function/src/index.ts
```diff
@@ -0,0 +1,57 @@
...
+    const fullPath = path.join(this.directory, url);
+    if (UP_PATH_REGEXP.test(fullPath)) {
+      return { action: 'sendStatus', status: 403 };
+    }
+    if (await isFile(fullPath)) {
+      return { action: 'serveFile', fullPath, contentType: mimeTypes.contentType(path.extname(fullPath)) };
+    }
+    let foundPath: string | undefined;
+    for (const ext of this.extensions) {
+      const p = `${fullPath + ext}`;
```
michaeladda: why do you need `${}` here? aren't they strings anyhow?

server-function/src/index.ts
```diff
@@ -0,0 +1,57 @@
...
+      if (await isFile(p)) {
+        foundPath = p;
+        break;
+      }
+    }
+    if (foundPath) {
+      return { action: 'serveFile', fullPath: foundPath, contentType: mimeTypes.contentType(path.extname(foundPath)) };
+    }
+    const notFoundPath = path.join(this.directory, '/404.html');
+    if (await isFile(notFoundPath)) {
```
michaeladda: not important: this can probably be tested once during initialization

server-function/src/index.ts
```diff
@@ -0,0 +1,57 @@
...
+  public async handle(url: string): Promise<any> {
+    if (url === '/invoke') {
+      return { action: 'handleInvoke' };
+    }
+    if (url === '/') {
+      return this.handle('/index.html');
+    }
+    const fullPath = path.join(this.directory, url);
+    if (UP_PATH_REGEXP.test(fullPath)) {
+      return { action: 'sendStatus', status: 403 };
```
michaeladda: most sites handle this by treating `..` of the root as `/` and would return 404 as usual if required.
for example: `../index.html` would just work if index.html exists.
i think you can mimic this with `path.join(this.directory, path.join('/', url))` and avoid the regexp completely

server-function/src/test/serve.test.ts
```diff
@@ -0,0 +1,11 @@
...
+import { Server } from '../index';
+
+async function testServer(t: ExecutionContext, url: string, expected: any) {
+  const server = new Server(process.cwd());
+  const value = await server.handle(url);
+  t.deepEqual(expected, value);
+}
+
+test('invoke', testServer, '/invoke', { action: 'handleInvoke' });
+test('not-found', testServer, '/invoke.jsx', { action: 'sendStatus', status: 404 });
```
michaeladda: i can think of a few more tests:
- `/404.html`
- `/` to `/index.html`
- up path
- auto `.html`

server-function/src/index.ts
```diff
@@ -0,0 +1,57 @@
...
+      return false;
+    }
+    throw err;
+  }
+}
+
+export class Server {
+  constructor(private directory: string, private extensions: string[] = ['.html']) {
+  }
+  public async handle(url: string): Promise<any> {
```
michaeladda: why any?
it's a pretty rigid type (action, sometimes status, sometimes fullPath and contentType)

server-function/src/index.ts
```diff
@@ -0,0 +1,57 @@
+import path from 'path';
+import fs from 'mz/fs';
```
michaeladda: not important: you can use fs/promises or util.promisify and save some deps

fetch-runtime/src/index.ts
```diff
@@ -14,7 +14,7 @@ export function createRuntime(methodNames: string[], fileId: Record<'filename',
     runtime[method] = (...args) => {
       let response: Response;
       // TODO: allow customizing URL
-      return fetch('/invoke', {
+      return fetch(process.env.PUBLIC_URL + '/invoke', {
```
bergundy: Do we need this once we support serving from `/invoke`?

server-function/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "@binaris/shift-server-function",
+  "version": "0.0.0",
```
bergundy: `0.0.1`?

server-function/shift_deploy_binaris.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
```
bergundy: I wish this was in typescript so I didn't need to rewrite it..

server-function/src/binaris.d.ts
```diff
@@ -0,0 +1,31 @@
+/// <reference types="node" />
```
bergundy: Maybe a comment on where this was copied from

server-function/src/index.ts
```diff
@@ -0,0 +1,106 @@
...
+import mimeTypes from 'mime-types';
+import ms from 'milliseconds';
+import fresh from 'fresh';
+
+// create-react-app uses only 32 bits for hashes, allow caching for a limited
+// time for lower chance of collision
+const CACHE_TIME_SEC = ms.months(1) / 1000;
+
+export async function isFile(filePath: string): Promise<fs.Stats | undefined> {
```
bergundy: Functions named `isX` usually return boolean, this is name is confusing for the functionality.

server-function/src/index.ts
```diff
@@ -0,0 +1,106 @@
...
+        contentType: mimeTypes.contentType('404.html'),
+        fullPath: notFoundPath,
+        cacheHint: {},
+        status: 404,
+      };
+    }
+    if (url === '/index.html') {
+      return { action: 'sendStatus', status: 404 };
+    }
+    return this.handle('/index.html', headers);
```
bergundy: Why is this the default?

server-function/src/index.ts
```diff
@@ -0,0 +1,106 @@
...
+    private directory: string,
+    private extensions: string[] = ['.html'],
+    private cachedPath = '/static',
+  ) {
+  }
+  public async handle(url: string, headers: { [k: string]: string | string[] | undefined }): Promise<Decision> {
+    if (url === '/invoke') {
+      return { action: 'handleInvoke' };
+    }
+    if (url === '/') {
```
bergundy: Do we plan on supporting `/subdir/` -> `/subdir/index.html`?

server-function/src/index.ts
```diff
@@ -0,0 +1,106 @@
...
+type Decision = ServeFile | Invoke | SendStatus;
+
+export class Server {
+  constructor(
+    private directory: string,
+    private extensions: string[] = ['.html'],
+    private cachedPath = '/static',
+  ) {
+  }
+  public async handle(url: string, headers: { [k: string]: string | string[] | undefined }): Promise<Decision> {
```
bergundy: This makes me think that we should have just made a custom runtime for shift functions implemented with express.
Would have been more robust, used standard tools and would have been easier to implement and test.

server-function/src/serveBinaris.ts
```diff
@@ -0,0 +1,60 @@
+import fs from 'mz/fs';
+import { Server } from './index';
+import { resolve as pathResolve } from 'path';
+const shiftServer = new Server('./build');
```
bergundy: Can this go under all of the imports please.

server-function/src/serveBinaris.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+export const handler: BinarisFunction = async (body, ctx) => {
+  const url = ctx.request.path;
+  const decision = await shiftServer.handle(url, ctx.request.headers);
+  switch (decision.action) {
+    case 'handleInvoke': {
+      if (!isInvokeRequest(body)) {
+        return new ctx.HTTPResponse({
+          statusCode: 400,
```
bergundy: Can you add a body explaining what's wrong with the request please?

server-function/shift_deploy_binaris.sh
```diff
@@ -0,0 +1,23 @@
...
+done
+cp -a `ls | grep -v node_modules | grep -v backend` .shift_deploy
+cat >.shift_deploy/binaris.yml <<EOF
+functions:
+  public_shift_builtin:
+    file: node_modules/@binaris/shift-server-function/dist/serveBinaris.js
+    entrypoint: handler
+    runtime: node8
+EOF
+./node_modules/.bin/babel --plugins @babel/plugin-transform-modules-commonjs backend/ -d .shift_deploy/backend
```
bergundy: Is the backend babel plugin not included here on purpose?

server-function/src/getdeps.ts
```diff
@@ -0,0 +1,26 @@
+import fs from 'fs';
+const contents = JSON.parse(fs.readFileSync('package-lock.json', 'utf8'));
```
bergundy: This fails if user does not have `package-lock.json` (I didn't), should we support this flow?

fetch-runtime/src/index.ts
vogre: Perhaps, to leave the possibility of multiple apps on same domain.

server-function/package.json
vogre: Will automatically roll to `0.0.1` when published.

server-function/shift_deploy_binaris.sh
vogre: Will be implemented separately, together will local-proxy support.

server-function/src/index.ts
vogre: Mostly agree, the cost this API is not in migration but breaking code reuse which assumes a framework.

server-function/src/index.ts
vogre: This is the default behavior of the webpack-dev-server. It does not appear in spec (lost somewhere in the comments). If we'll need something more complex we'll have no choice but allow the user to configure this (like https://firebase.google.com/docs/hosting/full-config#rewrites)

server-function/src/getdeps.ts
vogre: We can add a warning for several cases:

- user running the command outside of a CRA project (or a subdirectory).
- user chose to use yarn (and we don't support it yet).
- user manually removed package-lock.json and needs to run `npm install` before.

bergundy: This should be raised as an issue also linked to #104

PR #112, opened 2019-08-01
Generate client, server from interfaces

arielshaqed:
# ⚠️ This is an interim solution!
    
Rush and Concord do not play nicely together.  Create *fake* packages
for client and server definitions: `interfaces-{client,server}` link
during `rush update` as though they were the real thing, but after
build they forward to the generated code.
    
Problems include that dependencies from the generated code are not
propagated; we shall have to copy the really-required ones out from
`interfaces-{client,server}/gen/package.json` up to
`interfaces-{client,server}/package.json`.
    
# ⚠️ This is an interim solution!
    
The right way to do it may be to change Concord to support building a
package from inside it.

### DIFF
--- a/db/package.json
+++ b/db/package.json
@@ -15,7 +15,9 @@
   "author": "",
   "license": "MIT",
   "dependencies": {
-    "@binaris/shift-interfaces": "0.1.2",
+    "@binaris/shift-interfaces": "0.2.0-rc.0",
+    "@binaris/shift-interfaces-koa-server": "0.2.0-rc.0",
+    "@binaris/shift-interfaces-node-client": "0.2.0-rc.0",
     "@types/deep-freeze": "^0.1.1",
     "@types/leveldown": "^4.0.0",
     "@types/levelup": "^3.1.1",
--- a/db/src/db.ts
+++ b/db/src/db.ts
@@ -9,9 +9,12 @@ import {
   UpdateOptions,
   Version,
   Versioned,
-  Patch,
   KeyedPatches,
 } from '@binaris/shift-interfaces/dist/subscriptions';
+import {
+  Patch,
+  Serializable,
+} from '@binaris/shift-interfaces-node-client/gen/interfaces';
 import { ValueError } from './errors';
 import * as Q from './query';
 import { withTimeout, deferred, hrnano } from './utils';
@@ -22,14 +25,11 @@ export {
   Versioned,
   KeyedPatches,
   UpdateOptions,
+  Serializable,
 };
 
 const NUM_PATCHES_TO_KEEP = 20;
 const DEFAULT_READ_BLOCK_TIME_MS = 50000;
-
-// Typescript's way of defining any - undefined
-// see: https://github.com/Microsoft/TypeScript/issues/7648
-export type Serializable = {} | null;
 
 export function incrVersion([x, y]: Version, amount: number = 1): Version {
   return [x, y + amount];
--- a/db/src/db.ts
+++ b/db/src/db.ts
@@ -5,16 +5,16 @@ import { compare } from 'fast-json-patch';
 import LevelUpCtor, { LevelUp } from 'levelup';
 import LevelDown from 'leveldown';
 import { Mutex } from 'async-mutex';
+import { Version } from '@binaris/shift-interfaces/dist/db';
 import {
   UpdateOptions,
-  Version,
   Versioned,
   KeyedPatches,
 } from '@binaris/shift-interfaces/dist/subscriptions';
 import {
   Patch,
   Serializable,
-} from '@binaris/shift-interfaces-node-client/gen/interfaces';
+} from '@binaris/shift-interfaces-node-client/dist/interfaces';
 import { ValueError } from './errors';
 import * as Q from './query';
 import { withTimeout, deferred, hrnano } from './utils';
--- a/interfaces-client/package.json
+++ b/interfaces-client/package.json
@@ -1,0 +1,26 @@
+{
+  "name": "@binaris/shift-interfaces-node-client",
+  "private": true,
+  "version": "0.2.0-rc.0",
+  "files": [
+    "gen/"
+  ],
+  "description": "fake client interfaces wrapper",
+  "main": "gen/client.js",
+  "types": "gen/client.d.ts",
+  "scripts": {
+    "build": "npm run really-build",
+    "really-build": "rm -fr gen/ && concord 2>&1 --allow-extra-props node --client fetch -o gen ${npm_package_name}@${npm_package_version} ../interfaces/src/*.ts",
+    "lint": "echo generated client does not lint",
+    "test": "echo generated client needs no tests"
+  },
+  "devDependencies": {
+    "@binaris/shift-interfaces": "~0.2.0-rc.0",
+    "@types/node": "^10.14.9",
+    "concord": "~0.2.0",
+    "typescript": "^3.5.3"
+  },
+  "author": "Future Concord Piskie",
+  "repository": "undefined",
+  "license": "MIT"
+}
--- a/interfaces-client/package.json
+++ b/interfaces-client/package.json
@@ -3,21 +3,21 @@
   "private": true,
   "version": "0.2.0-rc.0",
   "files": [
-    "gen/"
+    "dist/"
   ],
   "description": "fake client interfaces wrapper",
-  "main": "gen/client.js",
-  "types": "gen/client.d.ts",
+  "main": "dist/client.js",
+  "types": "dist/client.d.ts",
   "scripts": {
     "build": "npm run really-build",
-    "really-build": "rm -fr gen/ && concord 2>&1 --allow-extra-props node --client fetch -o gen ${npm_package_name}@${npm_package_version} ../interfaces/src/*.ts",
+    "really-build": "rm -fr dist/ && concord 2>&1 --allow-extra-props node --client fetch -o dist ${npm_package_name}@${npm_package_version} ../interfaces/src/*.ts",
     "lint": "echo generated client does not lint",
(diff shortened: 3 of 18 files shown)


db/package.json
```diff
@@ -15,7 +15,9 @@
   "author": "",
   "license": "MIT",
   "dependencies": {
-    "@binaris/shift-interfaces": "0.1.2",
+    "@binaris/shift-interfaces": "0.2.0-rc.0",
```
bergundy: Is this still needed?

interfaces-client/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "scripts": {
+    "build": "npm run really-build",
+    "really-build": "rm -fr gen/ && concord 2>&1 --allow-extra-props node --client fetch -o gen ${npm_package_name}@${npm_package_version} ../interfaces/src/*.ts",
+    "lint": "echo generated client does not lint",
+    "test": "echo generated client needs no tests"
+  },
+  "devDependencies": {
+    "@binaris/shift-interfaces": "~0.2.0-rc.0",
+    "@types/node": "^10.14.9",
+    "concord": "~0.2.0",
```
bergundy: Can haz caret?

interfaces-server/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "version": "0.2.0-rc.0",
+  "files": [
+    "gen/"
+  ],
+  "description": "fake server interfaces wrapper",
+  "main": "gen/server.js",
+  "types": "gen/server.d.ts",
+  "scripts": {
+    "build": "npm run really-build",
+    "really-build": "rm -fr gen/ && concord 2>&1 --allow-extra-props node --server koa -o gen ${npm_package_name}@${npm_package_version} ../interfaces/src/*.ts",
```
bergundy: Server should not allow extra props.

interfaces/package.json
```diff
@@ -1,17 +1,19 @@
...
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "scripts": {
-    "build": "rm -rf dist/ && tsc && npm run compile-schema",
-    "compile-schema": "typescript-json-schema -o dist/schema.json --required --noExtraProps tsconfig.json '*'",
+    "build": "rm -rf dist/ && tsc",
     "lint": "tslint -c ../common/tslint.yml -p .",
-    "test": "echo interface needs no tests"
+    "test": "echo interface needs no tests",
+    "build-koa-server": "rm -fr gen-server/ && concord node --server koa -o gen-server ${npm_package_name}-koa-server@${npm_package_version} src/index.ts",
```
bergundy: Is this required here?

interfaces/package.json
```diff
@@ -1,17 +1,19 @@
...
-    "dist"
+    "dist/"
   ],
   "description": "interfaces for serving ShiftJS",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "scripts": {
-    "build": "rm -rf dist/ && tsc && npm run compile-schema",
-    "compile-schema": "typescript-json-schema -o dist/schema.json --required --noExtraProps tsconfig.json '*'",
+    "build": "rm -rf dist/ && tsc",
```
bergundy: Do we actually need to build this?

interfaces/src/db.ts
```diff
@@ -1,6 +1,12 @@
+import { Patch } from './subscriptions';
+
 export type Comparable = string | number | Date;
 export type Equatable = Comparable | boolean;
 
+// Typescript's way of defining any - undefined
+// see: https://github.com/Microsoft/TypeScript/issues/7648
+export type Serializable = {} | null;
```
bergundy: Does this generate the correct JSON schema?

interfaces/src/db.ts
```diff
@@ -87,3 +93,55 @@ export interface Query {
   skip?: number;
   orderBy?: ReadonlyArray<Order>;
 }
+
+export interface ClientContext {
+  debugId: string;
```
bergundy: Should `debugId` be generated by the server in this case?

interfaces/src/db.ts
```diff
@@ -87,3 +93,55 @@ export interface Query {
...
+export interface VersionedObject {
+  version: Version;
+  value: Serializable;
+}
+
+interface Patches {
+  /**
+   * Stores changes made to the document, meant to be used internally by poll().
+   */
+  patches: ReadonlyArray<Patch>;
```
bergundy: Drop the ReadOnly part please.

interfaces/src/db.ts
```diff
@@ -87,3 +93,55 @@ export interface Query {
...
+  get: {
+    params: { key: string; };
+    returns: Serializable | undefined;
+  };
+
+  /**
+   * Gets a single document with its version.
+   * @return - { version, value } or undefined if key doesn’t exist.
+   */
+  getWithMeta: {
```
bergundy: Not part of the public interface, should be `startPolling` (see `db/src/index.ts`).

interfaces/src/subscriptions.ts
```diff
@@ -1,5 +1,7 @@
 import { Operation } from 'fast-json-patch';
 
+export type Version = [number, number];
```
bergundy: Why is this redefined here?
Seems like we should simply unify the `subscriptions` and `db` sources.

rush.json
```diff
@@ -369,6 +369,16 @@
         "projectFolder": "interfaces",
         "shouldPublish": true
     },
+    {
+        "packageName": "@binaris/shift-interfaces-node-client",
+        "projectFolder": "interfaces-client",
+        "shouldPublish": false
```
bergundy: Why not?

subscriptions/package.json
```diff
@@ -15,7 +15,7 @@
   "license": "MIT",
   "dependencies": {
     "@binaris/shift-fetch-runtime": "0.0.3",
-    "@binaris/shift-interfaces": "0.1.2",
+    "@binaris/shift-interfaces": "0.2.0-rc.0",
```
bergundy: We should probably not reference this package directly anymore.

interfaces/src/db.ts
arielshaqed: This is a bit odd, that file wraps the DB interface to give the client.  E.g. even `update` etc. in that file don't take a `db` argument.  Not doing the rename until we better understand where to draw the line.

interfaces/src/subscriptions.ts
arielshaqed: Just reimporting for now.

Unifying the 2 files can come later.

rush.json
arielshaqed: I publish from `interfaces`.  Publishing from here would be odd, we would be using the *wrong* `package.json` -- the one in `gen/`, which (probably) means a different path from what we get this way.

subscriptions/package.json
arielshaqed: As above: we actually take from there.

interfaces/package.json
bergundy: Why? these are duplicates of the wrapper package scripts.
These commands don't publish either.

interfaces/src/db.ts
bergundy: Then it is not correct (you could call it a bug in typescript-json-schema).
We should probably just use `any` for the wire protocol.
If we can't  we could try to change the type to `{} | Array | string | number | boolean | null`.

interfaces/src/db.ts
bergundy: We never do this with external APIs, we always generate our own debugId and return it to the client.

interfaces/src/db.ts
```diff
@@ -87,3 +93,55 @@ export interface Query {
...
+}
+
+export interface DB {
+  /**
+   * Gets a single document.
+   * @return - value or undefined if key doesn’t exist.
+   */
+  get: {
+    params: { key: string; };
+    returns: Serializable | undefined;
```
bergundy: Maybe just define this as `any`?

interfaces/src/db.ts
```diff
@@ -87,3 +93,55 @@ export interface Query {
...
+  version: Version;
+  value: Serializable;
+}
+
+interface Patches {
+  /**
+   * Stores changes made to the document, meant to be used internally by poll().
+   */
+  patches: Patch[];
+  updatedAt: number;
```
bergundy: This isn't a part of patches IMO.

interfaces/src/db.ts
bergundy: We agreed we should use `startPolling` when we talked about this as this is the method that is supposed to signal the backend to prepare the subscription stream.

rush.json
bergundy: OK

db/src/db.ts
```diff
@@ -5,13 +5,16 @@ import { compare } from 'fast-json-patch';
...
   UpdateOptions,
-  Version,
   Versioned,
-  Patch,
   KeyedPatches,
 } from '@binaris/shift-interfaces/dist/subscriptions';
+import {
+  Patch,
+  Serializable,
+} from '@binaris/shift-interfaces-node-client/dist/interfaces';
```
bergundy: I understand why you did this but I don't like it.
I'd have preferred we have a pre-rush step to generate the required code: `npx concord@<version> --nocompile ..`.
I also understand that you're thinking of changing concord to fit into this situation, I thought about it a bit and it sounds like a legit option although I would still like to avoid it.

bergundy: Talking to @vogre I was convinced that modifying concord to not create a full package is a good option and should work well for us.

arielshaqed:
Closed for now in favour of #119.  It will be much less convenient, but hopefully get unstuck sooner.

PR #116, opened 2019-08-04
[db] Disallow all non-JSONable types at top level of values 

### LINKED ISSUE #115: [db] Don't allow non-JSONable values
Currently passing non-JSONable value (a function) causes item deletion.  Instead throw `ValueError`.

arielshaqed:
Fixes #115.

### DIFF
--- a/db/src/db.ts
+++ b/db/src/db.ts
@@ -67,9 +67,12 @@ export interface Document {
   value: Serializable;
 }
 
+// bigint not currently allowed.
+const allowedTypes = new Set(['object', 'boolean', 'number', 'string']);
+
 function checkValue(value: Serializable) {
-  if (typeof value === 'undefined') {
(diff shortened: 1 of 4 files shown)


db/src/db.ts
```diff
@@ -67,9 +67,12 @@ export interface Document {
   value: Serializable;
 }
 
+// bigint not currently allowed.
+const allowedTypes = new Set(['object', 'boolean', 'number', 'string']);
+
 function checkValue(value: Serializable) {
-  if (typeof value === 'undefined') {
-    throw new TypeError('Value must be not be undefined');
+  if (! allowedTypes.has(typeof value)) {
```
bergundy:
```suggestion
  if (!allowedTypes.has(typeof value)) {
```

arielshaqed:
Thanks!  Pulling once CircleCI spaces out.

PR #117, opened 2019-08-04
Rename shift-babel-macro to shift-frontend.macro

vogre:
To make imports mention macro only once

After publishing, the install script in https://github.com/binaris/shiftjs/blob/69d732ac66639ddf0d9db8985bb6f3ec01647c75/shiftjs-react-app/src/steps.ts#L51 needs to be updated.


michaeladda:
Also need to update the web app template and the example web apps.
in addition i'm not sure frontend and backend terminology is consistent with product specs.
please verify with @ashevat first
Anyhow, I think it would be better to rename everything once we migrate to the shiftjs org (when dropping the `shift-` prefix anyhow)

vogre:
Opened #118 as it's a common practice to have a babel-plugin and babel-macro in the same package even if their functionality differs.

vogre:
Went with #118 approach.

PR #119, opened 2019-08-05
Kluge Concord client, server generation

arielshaqed:
Use Concord to generate client and server packages.  These are never
stored in git, and must be generated whenever anything under
`interfaces/` changes.

Before even thinking of running `rush update`, run
`scripts/generate_interfaces`.  This generates directories
`gen-{client,server}`.  Rush immediately knows about them because they
are already mentioned in `rush.json`.  After updating, everything
should work normally.


interfaces/src/db.ts
```diff
@@ -79,11 +87,131 @@ export type Direction = 'ASC' | 'DESC';
...
+    params: { key: string; };
+    returns: any;
+  };
+
+  /**
+   * Gets a single document and its version.  First half of an update cycle.
+   */
+  getWithVersion: {
+    params: { key: string; };
+    returns: StoredDocument | Tombstone;
```
bergundy:
```typescript
{
  returns: { version: Version: value?: Serializable };
}
```

interfaces/src/db.ts
```diff
@@ -79,11 +87,131 @@ export type Direction = 'ASC' | 'DESC';
...
+    returns: boolean;
+  };
+
+  /**
+   * Gets the "current version" of a document with the intent to poll on it.
+   * @return - Versioned document or tombstone if it's marked deleted.
+   */
+  startPolling: {
+    params: { key: string; }
+    returns: StoredDocument | Tombstone;
```
bergundy: Same `returns` as `getWithVersion` please.

interfaces/src/db.ts
```diff
@@ -79,11 +87,131 @@ export type Direction = 'ASC' | 'DESC';
...
+ * A deleted document in the database, used for updating and for
+ * streaming.
+ */
+export interface Tombstone extends Patches {
+  updatedAt: number;
+  // (0, 0) if the document was never there in the first place.
+  version: Version;
+}
+
+export interface DB {
```
bergundy: I think all methods of this interface should include a table name parameter (we'll use default for now).

scripts/generate_interfaces
```diff
@@ -0,0 +1,15 @@
...
+set -euo pipefail
+
+(cd interfaces/ &&
+     npm install --no-package-json &&
+     npm run generate-koa-server && npm run generate-node-client)
+
+# Might not work if file already has "scripts" entry.
+function addScripts {
+    cp $1 $1.bk
+    jq  '. + { scripts: ["lint", "build", "test"] | map( { (.) : "echo generated code" } ) | add }' < $1.bk > $1
```
bergundy: It would have been better to run `tsc` on `build` (assuming concord was ran with `--nocompile`).

subscriptions/package.json
```diff
@@ -15,7 +15,7 @@
   "license": "MIT",
   "dependencies": {
     "@binaris/shift-fetch-runtime": "0.0.3",
-    "@binaris/shift-interfaces": "0.1.2",
+    "@binaris/shift-interfaces-node-client": "^0.2.0-rc.1",
```
bergundy:
```suggestion
    "@binaris/shift-interfaces-node-client": "^0.2.0",
```

interfaces/src/db.ts
bergundy: Let's discuss F2F.

arielshaqed: We did :-)
I explained why I don't know where to put it (client context?  parameter??).  It depends on understanding the feature, which in turns depends on designing it. So not done for now.

PR #124, opened 2019-08-07
Avoid creating interfaces/package-lock.json

vogre:
Which needs to be manually removed each time

### DIFF
--- a/scripts/generate_interfaces
+++ b/scripts/generate_interfaces
@@ -2,7 +2,7 @@
 set -euo pipefail
 
 (cd interfaces/ &&
-     npm install --no-package-json &&
+     npm install --no-package-lock &&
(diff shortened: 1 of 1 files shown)


scripts/generate_interfaces
```diff
@@ -2,7 +2,7 @@
 set -euo pipefail
 
 (cd interfaces/ &&
-     npm install --no-package-json &&
+     npm install --no-package-lock &&
```
arielshaqed: Whaaa??!  Thanks for the fix, it's obviously great.  But why no error message?
```
flutterby:thoth ariels$ npm install --no-way-v6.9.0-of-a-tool-can-be-this-broken-not-even-npm
audited 4391 packages in 5.117s
found 0 vulnerabilities
```

# 🤢

PR #128, opened 2019-08-08
Restore missing macro.js file

vogre:
Needed for pretty imports of 'code-transform/macro'.


vogre:
Tested by manually packaging with `$ rush publish --include-all --pack -p`.

PR #131, opened 2019-08-08
Re-publish shift CLI

vogre:
For a pretty 0.0.1 version


vogre:
A republish is not a must, it would only affect the version being shown as 0.0.1 as opposed to unfamiliar 0.0.0. Can wait until real features added to CLI.

vogre:
Obsolete, since shift-cli published with `0.1.0`.

PR #134, opened 2019-08-11
Shift CLI Authentication Flow

arikmaor:
This PR adds a custom base class for all the cli commands

The base command:
1. Adds default flags like `--help` and `--realm`
2. Initializes a `LycanClient` instance to be used inside commands with 'this.lycanClient'
3. Adds a `async authenticate(force = false)` method to the command subclasses which runs the authentication flow, saves the accessToken and returns it (or just returns it if it's saved)


cli/src/utils/user-config.ts
```diff
@@ -0,0 +1,13 @@
+import { join } from 'path';
+import { homedir } from 'os';
+import Conf from 'conf';
+import { safeDump, safeLoad } from 'js-yaml';
+
+export default new Conf({
+  fileExtension: 'yml',
```
bergundy: Why do we care if it's yaml if we don't intend for it to be edited manually?
Is this part of the spec?

cli/src/utils/user-config.ts
```diff
@@ -0,0 +1,13 @@
...
+  serialize: safeDump,
+  deserialize: safeLoad,
+  clearInvalidConfig: false,
+  configName: 'shiftjs.config',
```
bergundy: Just call this `config` if it's already in `~/.shiftjs`.

cli/src/utils/command.ts
```diff
@@ -0,0 +1,114 @@
...
+
+export default abstract class BaseCommand extends Command {
+  static cliBinName = pjson.oclif.bin as string;
+
+  protected realm?: string;
+  protected lycanClient?: LycanClient;
+
+  public static flags = {
+    help: flags.help({ char: 'h' }),
+    realm: flags.string({
```
bergundy: Can we drop this please?
`realm` is binaris internal.
I'd rather just have a base URL, it's much more generic.

cli/src/utils/command.ts
```diff
@@ -0,0 +1,114 @@
...
+    return require('@oclif/parser').parse(argv, { context: this, ...opt });
+  }
+
+  public async init() {
+    const { flags: { realm } } = this.parse(BaseCommand);
+    this.realm = realm;
+    this.lycanClient = new LycanClient(`${this.getBaseUrl('api')}/public/v1`);
+  }
+
+  public async authenticate(force = false): Promise<string> {
```
bergundy: `force` is not a good name, can we call it `useLocalToken` or `skipLocalToken` or `remoteOnly`?

cli/src/utils/command.ts
```diff
@@ -0,0 +1,114 @@
...
+    return accessToken;
+  }
+
+  private getBrowserLoginUrl(ticket: string): string {
+    const loginUrl = new URL(`${this.getBaseUrl('app')}/cli-login`);
+    loginUrl.searchParams.set(LOGIN_PARAM, ticket);
+    return loginUrl.href;
+  }
+
+  private getBaseUrl(subdomain: string, protocol: 'http' | 'https' = 'https') {
```
bergundy: As mentioned before, I'd rather just have 2 config variables or environment variables for the api URL and website URL.

bergundy: For example, how do I develop locally with only realm as a configuration value?

cli/src/utils/command.ts
```diff
@@ -0,0 +1,114 @@
...
+      opt!.flags = BaseCommand.flags as any;
+    }
+
+    return require('@oclif/parser').parse(argv, { context: this, ...opt });
+  }
+
+  public async init() {
+    const { flags: { realm } } = this.parse(BaseCommand);
+    this.realm = realm;
+    this.lycanClient = new LycanClient(`${this.getBaseUrl('api')}/public/v1`);
```
bergundy: The client should be initialized with the `shift-api-key` header if an API key is stored in the config..

cli/src/utils/user-config.ts
arikmaor: yes
> When the user successfully completes the process, logic will create a credentials file at `~/.shiftjs/shiftjs.config.yml` which will store the authenticated credentials.

cli/src/utils/user-config.ts
arikmaor: from the spec:
> When the user successfully completes the process, logic will create a credentials file at `~/.shiftjs/shiftjs.config.yml` which will store the authenticated credentials.

cli/src/utils/command.ts
```diff
@@ -0,0 +1,114 @@
...
+const pjson = JSON.parse(readFileSync(resolve(__dirname, '../../package.json'), 'utf-8'));
+
+const LOGIN_PARAM = 'ticket';
+const TICKET_CLAIM_INTERVAL_MS = 1000;
+
+export default abstract class BaseCommand extends Command {
+  static cliBinName = pjson.oclif.bin as string;
+
+  protected realm?: string;
+  protected lycanClient?: LycanClient;
```
bergundy: Would be more useful to have a `getLycanClient` method than to check if lycanClient is defined every time you want to use it.

cli/src/utils/command.ts
arikmaor: just to clarify, realm is not printed to the user, it's hidden and visible only inside the code
for example, do you want to ignore the realm environment variable completely?

cli/src/utils/command.ts
arikmaor: `useLocalToken` - I don't like it because you usually use it.
`skipLocalToken` - Not a good name IMO because the new token will be saved
`remoteOnly` - Doesn't say anything to me

do you have a problem with the world `force`?
how about 'forceBrowserAuthFlow' ?

cli/src/utils/command.ts
arikmaor: done

cli/src/utils/command.ts
arikmaor: done

cli/src/utils/command.ts
arikmaor: done

cli/src/utils/command.ts
arikmaor: done (used property instead)

cli/src/utils/command.ts
bergundy: forceBrowserAuthFlow is good, thanks!

cli/src/utils/command.ts
```diff
@@ -0,0 +1,120 @@
...
+    return loginUrl.href;
+  }
+
+  private createLycanClient(apiKey?: string): LycanClient {
+    const options = apiKey ? {
+      headers: {
+        'shift-api-key': apiKey,
+      },
+    } : {};
+    return new LycanClient(`${this.apiEndpoint}/public/v1`, options);
```
bergundy: `/public/v1` could be part of the URL.

cli/src/utils/command.ts
```diff
@@ -0,0 +1,120 @@
...
+  }
+}
+
+function sleep(durationMs: number) {
```
bergundy: Can we put this in utils please.

cli/src/utils/command.ts
```diff
@@ -0,0 +1,120 @@
...
+      hidden: true,
+      env: 'SHIFTJS_WEBAPP_ENDPOINT',
+    }),
+  };
+
+  protected parse<F, A extends {[name: string]: any}>(opt?: Parser.Input<F>, argv = this.argv): Parser.Output<F, A> {
+    if (!opt) {
+      opt = this.constructor as any;
+    }
+    if (!opt!.flags) {
```
bergundy: Do we still need the `!`?

cli/src/utils/command.ts
```diff
@@ -0,0 +1,120 @@
...
+      opt!.flags = BaseCommand.flags as any;
+    }
+
+    return require('@oclif/parser').parse(argv, { context: this, ...opt });
```
bergundy: Can we not require dynamically?

cli/src/utils/command.ts
arikmaor: done

cli/src/utils/command.ts
arikmaor: done (and also the other one)

cli/src/utils/command.ts
arikmaor: done

cli/src/utils/command.ts
arikmaor: yes, but I think it never happens, I can add a check/throw if you want

cli/src/utils/command.ts
arikmaor: done

cli/src/utils/command.ts
arikmaor: changed how the typing here is handled

cli/src/utils/command.ts
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  private createLycanClient(apiKey?: string): LycanClient {
+    const options = apiKey ? {
+      headers: {
+        'shift-api-key': apiKey,
+      },
+    } : {};
+    if (!this.apiEndpoint) {
+      throw new Error('apiEndpoint not set!');
```
bergundy: Should this be `this.error()`?

cli/src/utils/command.ts
```diff
@@ -0,0 +1,123 @@
...
+
+    this.log('Successfully logged into ShiftJS!');
+    userConfig.set({ accessToken });
+    this._lycanClient = this.createLycanClient(accessToken);
+    return accessToken;
+  }
+
+  private getBrowserLoginUrl(ticket: string): string {
+    if (!this.webAppLoginUrl) {
+      throw new Error('webAppLoginUrl not set!');
```
bergundy: Should this be `this.error()`?

PR #135, opened 2019-08-12
Re-shape DB *Context interfaces

arielshaqed:
- Stop saying "token", it's just confusing.
- Authenticate using apiKey.
- Specify _what_ we're doing with appId.
- Also add server-side specified debugId.

These are pretty much already present server-side.


subscriptions/package.json
```diff
@@ -15,7 +15,7 @@
   "license": "MIT",
   "dependencies": {
     "@binaris/shift-fetch-runtime": "0.0.3",
-    "@binaris/shift-interfaces-node-client": "^0.2.0",
+    "@binaris/shift-interfaces-node-client": "^0.3.0",
```
vogre: Note that since the concord package needs to be managed manually and not by rush, a `rush change` should be generated.

To emulate rush the patch message should be "Version update only".

arielshaqed: Yeah, I figured.  Anyway it does deserve a rush change (it's a very different package now).

An annoyance is that the actual changed packages -- `shift-interfaces-*-{client,server}` get no changes because they're generated.

PR #141, opened 2019-08-13
Use cli-ux for waiting, linking and spinners

arikmaor:


common/changes/shift-cli/cli-spinner_2019-08-13-11-37.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "comment": "",
```
vogre: Note that an empty changelog will not get the changes to the package published until the next non-empty changelog.

arikmaor: ack

arikmaor:
@arielshaqed 
> How was this tested?
Manually, but I'm working right now no adding a unit tests framework to the cli

> Does it work on non-iTerms?
Spinner works in osx terminal
Links are displayed like this: "linktext (https://url.you.linked.to.com)" instead of "linktext" with underline that can be clicked

arielshaqed:
> @arielshaqed
> 
> > How was this tested?
> 
> Manually, but I'm working right now no adding a unit tests framework to the cli

Sure, but that probably won't work for making sure it looks good on a terminal.  This is a case where manually testing (even just once) helps more than automated tests.

> 
> > Does it work on non-iTerms?
> 
> Spinner works in osx terminal
> Links are displayed like this: "linktext (https://url.you.linked.to.com)" instead of "linktext" with underline that can be clicked

Cool.  I see in the code that it also tries to work for WIN32.  That leaves Linux terminals.  It's OK to degrade gracefully (e.g. show the classic `|\-/|\-/` spinner, but less OK if it fills the screen with garbage.  Terminals are so much fun!

Reading [the code](https://github.com/oclif/cli-ux/blob/611b4d8b720b93abb33fc00e235a7e820911412c/src/config.ts#L18) it seems to try to set up a good spinner.  You might check it by setting `TERM` to interesting values, maybe `export TERM=vt100` or `export TERM=xterm`.  These should give you an ASCII spinner, not the nice MacOS spinning dots.

arielshaqed:
Looking  a bit further, it should work with any Unicode-capable terminal.  So good enough for now!

Thanks!

PR #143, opened 2019-08-13
Implement logs

bergundy:


michaeladda:
tested manually. will add CLI tests once their framework is complete

cli/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "shift-cli",
   "description": "ShiftJS CLI Tool",
-  "version": "0.1.2",
+  "version": "0.1.3",
```
vogre: This is probably wrong, if there is a `rush change` of type `patch` it will automatically increment the version to `0.1.3` on publish.

cli/src/commands/logs.ts
```diff
@@ -25,18 +31,19 @@ $ ${Command.cliBinName} logs --since 2m --follow`,
 
   public static flags = {
     ...Command.flags,
-    limit: flags.minMaxInt({
+    limit: flags.integer({
```
vogre: Can we go back to `minMaxInt`?

cli/src/commands/logs.ts
```diff
@@ -25,18 +31,19 @@ $ ${Command.cliBinName} logs --since 2m --follow`,
...
       char: 'l',
       description: 'Limit number of entries shown (cannot exceed 1000).',
       default: 500,
-      min: 1,
-      max: 1000,
+      // min: 1,
```
arielshaqed: Why?  Didn't we just add documentation with `min`, `max` on https://github.com/binaris/shiftjs/pull/143/files#diff-ccac2b54f1b9872d1d8933d39a4514d1R80?

cli/src/commands/logs.ts
```diff
@@ -48,10 +55,27 @@ $ ${Command.cliBinName} logs --since 2m --follow`,
   public static strict = true;
 
   public async run() {
+    const { since, follow, limit } = this.parse(Logs).flags;
```
arielshaqed: Why change the original destructuring syntax?  (I don't care one way or the other particularly, just seems strange to switch).

cli/src/commands/logs.ts
```diff
@@ -48,10 +55,27 @@ $ ${Command.cliBinName} logs --since 2m --follow`,
...
+    await this.authenticate();
 
-    const {flags: { since, follow, limit }} = this.parse(Logs);
-
-    this.log('Logs', { since, follow, limit });
+    const projectDir = await getProjectRootDir();
+    const projects = conf.get('projects') as Project[] | undefined || [];
```
arielshaqed: This might be simpler as
```suggestion
    const projects = (conf.get('projects') || []) as Project[];
```
It should certainly emit the same code.

cli/src/utils/cli-flags.ts
```diff
@@ -10,24 +10,25 @@ export type MinMaxIntFlagOptions = Parameters<(typeof flags.integer)>[0] & {
...
-        return val;
+        return new Date(val);
       }
       const valMs = ms(val);
       if (valMs !== undefined) {
-        return valMs;
+        return val;
       }
-      error(`Expected a formatted duration format but received: ${val}`);
+      error(`Expected a formatted duration or isISO8601 string format but received: ${val}`);
```
vogre: `isISO8601` -> `ISO 8601`.

cli/src/commands/logs.ts
```diff
@@ -48,10 +55,27 @@ $ ${Command.cliBinName} logs --since 2m --follow`,
...
+    const project = projects.find(({ directory }) => directory === projectDir);
+    if (project === undefined) {
+      return this.error(`No project deployments found, please run ${Command.cliBinName} deploy`);
+    }
+    let token;
+    do {
+      // TODO: support other envs
+      const sinceDate = typeof since === 'string' ? new Date(Date.now() - ms(since)) : since;
+      const { records, nextToken } = await this.lycanClient.getLogs(
+        project.applicationId.replace(/-/g, ''), project.defaultEnv, { follow, limit, since: sinceDate });
```
vogre: Perhaps need to do it on the server, since this somewhat guarantees that `a-b-c` and `ab-c` are the same app...

cli/src/commands/logs.ts
michaeladda: we dropped it until minMaxInt will support defaults properly

cli/src/commands/logs.ts
michaeladda: it needs typing improvement first :-/

cli/src/commands/logs.ts
michaeladda: I suspect for readability. I, too, like Roey find the new format more readable.

cli/src/commands/logs.ts
michaeladda: 👍

michaeladda: sorry. simpler, yes, but not good due to linting

michaeladda:
Thanks, and indeed we didn't define the shiftjs conf type. will address it in another PR.

PR #146, opened 2019-08-14
Change local invoke-server to use Koa

arielshaqed:
This is the *inner* server, which runs user backend code.  Change
it to use Koa.  This lets us embed Concord-based servers in it.

The local proxy itself has to be based on Express.  This is required
by CRA.

Tested (only) by single local unit-test.


local-proxy/src/server.ts
```diff
@@ -35,7 +36,9 @@ const whitelistedModules = new Map(
   }).filter((pair): pair is [string, any] => pair !== undefined)
 );
 
-const app = express();
+const app = new Koa();
+app.use(bodyParser({ enableTypes: ['json'], strict: false, detectJSON: () => true }));
```
vogre: I think no need for `detectJSON`, as all requests are generated by browser and no need for manual testing with cURL where omitting content-type header makes it easier.

We can also assume `strict: true`, since the requests are always sending objects - but this seems less critical since we verify the parameters anyway.

local-proxy/src/server.ts
```diff
@@ -102,14 +108,18 @@ app.post('/invoke', json(), async (req, res) => {
     // Not logging args to avoid sensitive info log (?)
     reportInvocationDurationUs(startHrtime, requestId, { path, handler });
     if (ret === undefined) {
-      return res.sendStatus(204);
+      ctx.status = 204;
+      return;
     }
-    res.json(ret);
+    // ret might be a scalar, which Koa does not auto-JSONify.  So
```
vogre: Nice catch.

local-proxy/src/server.ts
```diff
@@ -73,14 +76,15 @@ function reportInvocationDurationUs(startTime: [number, number], requestId: stri
   return durationMicro;
 }
 
-app.post('/invoke', json(), async (req, res) => {
-  if (req.headers['x-shift-dev-server-local-token'] !== localToken) {
-    return res.sendStatus(403);
+router.post('/invoke', async (ctx, _next) => {
```
vogre: `_next` can be omitted if unused.

arielshaqed: True.  Done.

local-proxy/src/server.ts
arielshaqed: The unit test returns a scalar value (as it should, those are always the hardest).

arielshaqed:
Thanks!

PR #148, opened 2019-08-18
Add DB client and local DB server

arielshaqed:
## ⚠️ WIP: Safe to merge (passes tests) but not yet hooked up! ⚠️ 

1. *Local DB server.*  This is based on the local DB backend implementation, in particular it stores in `leveldb`.  Not yet hooked up to the local proxy (upcoming PR).
2. *Bugfixes to DB Concord interface.*  As part of splitting out the DB I discovered some bugs in my interface, and changed the interface accordingly.  This has no b/c effects as we have no released DB clients.  I will port these changes to Thoth as part of integration.
3. *DB client.*  Based on above interface.

There are some server tests.  They mostly replicate the previous DB tests.  We do not yet have integration tests, and the client remains untested.


db-client/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@binaris/shift-db-client",
+  "version": "0.0.1",
+  "description": "(Backend) client for accessing ShiftJS DBs.",
+  "main": "dist/index.js",
```
vogre: Possibly, a files section including only dist.

db-client/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@binaris/shift-db-client",
+  "version": "0.0.1",
```
vogre: 0.0.0 for publish to bump it.

db/src/db.ts
```diff
@@ -131,7 +131,7 @@ export class DB extends EventEmitter {
       const val = await this.db.get(key);
       return JSON.parse(val.toString());
     } catch (err) {
-      if (err.name === 'NotFoundError') {
+      if (err.notFound) {
```
vogre: Why is this change needed if db intended to be removed? A different `getWithMeta` uses `NotFoundError`...

interfaces/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@binaris/shift-interfaces",
-  "version": "0.3.0",
+  "version": "0.4.0",
```
vogre: Should all the dependencies (`db`, `subscriptions`) be updated to use this version?

common/changes/@binaris/shift-leveldb-server/split-db_2019-08-18-08-47.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@binaris/shift-leveldb-server",
+      "comment": "New server to add to local-proxy",
+      "type": "major"
```
vogre: patch for 0.0.1?

leveldb-server/src/db.ts
```diff
@@ -67,6 +67,10 @@ function versionsMatch(prev: StoredDocument | Tombstone | undefined, version: Ve
   return prev.version.major === version.major && prev.version.minor === version.minor;
 }
 
+function valueOrUndefined(maybeHasValue?: any) {
+  return maybeHasValue === undefined ? undefined : maybeHasValue.value!;
```
vogre: Can it be `value` without `!`?

db-client/src/index.ts
```diff
@@ -0,0 +1,160 @@
...
+function serverUrlFor(endpoint: string): string {
+  return `https://${endpoint}/v1`;
+}
+
+const defaultOptions: Options = {
+  timeoutMs: 2000,
+};
+
+// Not an async iterator so as to separate *generation* of the next
+// delay from its *activation*.
```
vogre: Note that the delay is activated immediately, it is only awaited at a random time in the future.

db-client/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "../common/tsconfig.json",
+  "compilerOptions": {
+    "outDir": "dist",
+    "downlevelIteration": true
```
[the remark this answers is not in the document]
arielshaqed: Also I don't really have async iteration in this code :-/

db/src/db.ts
arielshaqed: True.  This is actually the even better way to do it, so changing _that_ one.

interfaces/package.json
arielshaqed: Not yet.  `db` is going away but not yet.  `subscriptions` is even further away, for now.

leveldb-server/CHANGELOG.json
```diff
@@ -0,0 +1,171 @@
+{
```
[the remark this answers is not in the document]
arielshaqed: Yup.  Also the MD I copied blindly.

leveldb-server/src/db.ts
arielshaqed: Done.  Leftover from older longer code.

.gitignore
```diff
@@ -64,6 +64,8 @@ dist/
 common/temp/**
 package-deps.json
 
+package-lock.json
```
vogre: Maybe unneeded with #151.

leveldb-server/src/test/db.match.spec.ts
```diff
@@ -0,0 +1,175 @@
+import test, { Macro } from 'ava';
+import { testing } from '../db';
+import { Filter } from '@binaris/shift-interfaces-koa-server/interfaces';
+
+const { match } = testing;
+
+type Operator = 'eq' | 'ne' | 'gt' | 'lt' | 'gte' | 'lte';
```
michaeladda: maybe generated this directly in `interfaces/src/db.ts`?

leveldb-server/src/test/db.match.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+  // creates a legal filter, so just cast.
+  const filter = { operator, path: ['key'], value: matchKey } as Filter;
+  t.false(match({ key: testKey, value: 17 }, filter));
+};
+keyDoesntMatch.title = (providedTitle = '', testKey, operator, matchKey) =>
+  `${providedTitle} !${operator}(key ${JSON.stringify(testKey)}, key ${JSON.stringify(matchKey)})`;
+
+test('eq checks for strict equality', valueMatches, 3, 'eq', 3);
+test('eq checks for strict equality', valueDoesntMatch, 3, 'eq', 2);
+test('eq checks for strict equality', valueDoesntMatch, 3, 'eq', '3');
```
michaeladda: nice!
I'd test null and 'null', undefined and 'undefined' too since it's interesting

leveldb-server/src/test/db.match.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+  const caseInsensitive = pattern.flags.includes('i');
+  // TypeScript type system not strong enough to deduce this always
+  // creates a legal filter, so just cast.
+  const filter = { operator: 'matches' as 'matches', path: ['value'], pattern: pattern.source, caseInsensitive };
+  t.false(match({ key: 'abc', value: testValue }, filter));
+};
+stringDoesntMatch.title = (providedTitle = '', testValue, matchValue) =>
+  `${providedTitle} ${JSON.stringify(testValue)} !~ ${matchValue})`;
+
+test('matches matches string', stringDoesntMatch, null, /abc/i);
```
michaeladda: also match undefined?

leveldb-server/src/test/db.match.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+  `${providedTitle} ${testValue}.startsWith(${value})`;
+
+const doesntStartWith: Macro<[any, string]> = (t, testValue, value) => {
+  const filter = { operator: 'startsWith' as 'startsWith', path: ['value'], value };
+  t.false(match({ key: 'abc', value: testValue }, filter));
+};
+doesntStartWith.title = (providedTitle = '', testValue, value) =>
+  `${providedTitle} !${testValue}.startsWith(${value})`;
+
+test('startsWith matches string', doesntStartWith, null, 'ab');
```
michaeladda: also match undefined?

leveldb-server/src/test/db.match.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+test('startsWith matches string', doesntStartWith, 0, 'ab');
+test('startsWith matches string', startsWith, 'abc', 'ab');
+test('startsWith matches string', doesntStartWith, 'ABC', 'ab');
+
+const startsWithAb = { operator: 'startsWith' as 'startsWith', path: ['key'], value: 'ab' };
```
michaeladda: using value `Ab` might be more readable

leveldb-server/src/test/db.match.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+const equalsAb = { operator: 'eq' as 'eq', path: ['key'], value: 'ab' };
+const equals0 = { operator: 'eq' as 'eq', path: ['value'], value: 0 };
+const equals1 = { operator: 'eq' as 'eq', path: ['value'], value: 1 };
+
+test('and applies all filters', (t) => {
+  t.true(match({ key: 'abc', value: 0 }, { operator: 'and', filters: [startsWithAb, equals0] }));
+  t.false(match({ key: 'abc', value: 0 }, { operator: 'and', filters: [startsWithAb, equals1] }));
+  t.false(match({ key: 'abc', value: 0 }, { operator: 'and', filters: [equalsAb, equals0] }));
+  t.false(match({ key: 'abc', value: 0 }, { operator: 'and', filters: [equalsAb, equals1] }));
```
michaeladda: consider adding match with filter of same path
`t.true(match({ key: 'ab', value: 0 }, { operator: 'and', filters: [startsWithAb, equalsAb] }));`
and
`t.true(match({ key: 'abc', value: 0 }, { operator: 'and', filters: [startsWithAb, equalsAb] }));`

db-client/src/index.ts
```diff
@@ -0,0 +1,162 @@
...
+      v1: {
+        // TODO(ariels): Update with prefixes from binaris/spice#616.
+        appId: process.env.APP_ID!,
+        apiKey: process.env.API_KEY!,
+      },
+    },
+  };
+
+  constructor(options?: Options) {
+    this.client = new DBClient(serverUrlFor(process.env.DB_ENDPOINT!), options || defaultOptions);
```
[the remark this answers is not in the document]
arielshaqed: Good idea, done!

db-client/src/index.ts
```diff
@@ -0,0 +1,162 @@
...
+  public create(key: string, value: Serializable): Promise<boolean> {
+    return this.client.create(this.ctx, key, value);
+  }
+
+  public remove(key: string): Promise<boolean> {
+    return this.client.remove(this.ctx, key);
+  }
+
+  // TODO(ariels): Support operationId for streaming.
+  public async update<T extends Serializable = any>(
```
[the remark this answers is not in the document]
arielshaqed: `T` will typically be derived from the type of `updater`.  I'm not sure when a non-`any` default will be useful (AFAIK TypeScript will _not_ be able to force a default type onto a generic `updater`).

db-client/src/index.ts
```diff
@@ -0,0 +1,162 @@
...
+    key: string, updater: (state?: DeepReadonly<T>) => T, _options?: UpdateOptions,
+  ): Promise<DeepReadonly<T>> {
+    for (const delay of backoff()) {
+      const { value, version } = await this.getWithVersion(key);
+      const newValue = updater(deepFreeze(value as T));
```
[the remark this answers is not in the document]
arielshaqed: Same as existing code, no?

Anyway binaris/shiftjs-beta-site#9

db-client/src/index.ts
```diff
@@ -0,0 +1,162 @@
...
+    throw new Error('Unimplemented');
+  }
+
+  public async find(): Promise<any[]> {
+    throw new Error('Unimplemented');
+  }
+
+  private setIfVersion(
+    key: string,
+    value: {} | any[] | null | string | number | boolean,
```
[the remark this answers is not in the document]
arielshaqed: Done (an advantage of using the same `Serializable`).

db-client/src/index.ts
```diff
@@ -0,0 +1,162 @@
...
+}
+
+const db = new DBHandler();
+
+/**
+ * Gets a single document.
+ * @return - value or undefined if key doesn’t exist.
+ */
+export async function get<T extends Serializable = any>(key: string): Promise<T | undefined> {
+  return await db.get(key);
```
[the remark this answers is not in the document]
arielshaqed: Done.

db-client/src/index.ts
```diff
@@ -0,0 +1,162 @@
...
+  return await db.remove(key);
+}
+
+/**
+ * Updates a single document.
+ * @param updater - Function that gets the previous value and returns the next value to update the DB with.
+ *                  Cannot return undefined, receives undefined in case key doesn’t already exist in the DB.
+ * @return - The new value returned from updater
+ */
+export async function update<T extends Serializable = any>(
```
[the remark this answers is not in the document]
arielshaqed: Same as before.

db-client/src/index.ts
```diff
@@ -0,0 +1,162 @@
...
+  throw new Error('Unimplemented');
+}
+
+/**
+ * Polls on updates to specified keys since specified versions.
+ */
+export async function poll(): Promise<any> {
+  throw new Error('Unimplemented');
+}
+poll.__shiftjs__ = { exposed: true };
```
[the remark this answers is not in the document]
arielshaqed: Missed in copy.

db-client/src/index.ts
```diff
@@ -0,0 +1,162 @@
...
+
+/**
+ * Gets a initial document in an intent to for poll on it.
+ */
+export async function startPolling<T extends Serializable = any>(_key: string): Promise<Versioned<T | undefined>> {
```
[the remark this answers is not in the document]
arielshaqed: It will be there in future, when I add polling.  It's already present in `db`, this avoids breaking existing code.

interfaces/src/db.ts
```diff
@@ -160,6 +165,8 @@ interface Patches {
   patches: Patch[];
 }
 
+export type KeyedVersions = Array<[string, Version]>;
```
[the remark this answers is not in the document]
arielshaqed: But `type` is only an alias, not a type (so like `typedef` in C and especially in C++, and not at all like `type` in Go).  So we'll have confusing error messages regarding this type, sometimes it will be `Key` and sometimes it will be `string`.  Given that `Key` is such a useful name, I'd rather not.

leveldb-server/src/db.ts
```diff
@@ -104,50 +99,63 @@ export class DB extends EventEmitter {
...
    */
-  public async get<T extends Serializable = any>(key: string): Promise<T | undefined> {
-    const versioned = await this.getWithMeta(key);
+  public async get(ctx: ServerOnlyContext, key: string): Promise<Serializable | undefined> {
+    const versioned = await this.getWithMeta(ctx, key);
     if (versioned === undefined) {
       return undefined;
     }
-    return versioned.value;
+    return (versioned as any).value;
```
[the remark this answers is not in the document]
arielshaqed: `versioned` might be a `Tombstone`.  In that case it has no `value` field but I still want to get `undefined`.  Tombstone cannot have a `value: undefined`, because that isn't JSONable, and we'll _anyway_ get it with no `value` field.

leveldb-server/src/test/db.match.spec.ts
arielshaqed: Not done.
1. I'd need to break up `Filter` into a union of multiple sub-types, and then take `SubFilter['operator']`.  That break-up would have to go according to what is convenient for this test, not necessarily making sense elsewhere.
2. If a new operator is added, I _anyway_ have to write tests for it.  So I can expand `Operator` at that time.  This is _not_ a type that can just expand because it passes through.

leveldb-server/src/test/db.match.spec.ts
arielshaqed: Same tests as we had for `db`.

We cannot test `undefined` because that's not a DB value (not JSONable).  And this test doesn't check every type combination; rather, it asserts that we are using `===` and not `==`.  Adding more cases will not make it stricter.

leveldb-server/src/test/db.match.spec.ts
arielshaqed: As above: `undefined` is not JSONable.

leveldb-server/src/test/db.match.spec.ts
arielshaqed: As above.

leveldb-server/src/test/db.match.spec.ts
arielshaqed: I find @bergundy 's version (mostly copied above) to be easier to read, because the 2 parts are quite different.

db-client/src/index.ts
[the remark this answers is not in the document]
arielshaqed: OK, I'll do it.  I don't care which style we pick, so I'll go with yours.

db-client/src/index.ts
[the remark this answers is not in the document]
arielshaqed: True, returning to exposing only those 2.  We will have the backend authorization discussion separately.

db-client/src/index.ts
```diff
@@ -0,0 +1,164 @@
...
+import deepFreeze, { DeepReadonly } from 'deep-freeze';
+import { merge } from 'ramda';
+
+export interface Versioned<T extends Serializable | undefined> {
+  version: Version;
+  value: T;
+}
+
+function serverUrlFor(endpoint: string): string {
+  return `https://${endpoint}/v1`;
```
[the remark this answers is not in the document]
arielshaqed: True.  It will just be a _base_ URL, though -- which `/v<number>` you go to depends on the client.

PR #150, opened 2019-08-19
Disable package lock

michaeladda:
manually tested

### DIFF
--- a/cli/.npmrc
+++ b/cli/.npmrc
@@ -1,0 +1,1 @@
+package-lock=false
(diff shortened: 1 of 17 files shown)


vogre:
I don't think this is right, as this is very much like `.gitignore`, it hides the issue of running `npm install` which might clobber the symlinked `node_modules`, and not prohibits `npm install` completely (and tells you something was wrong).

The `package-lock.json` file itself is harmless, it's the effect on `node_modules` that running `npm` commands that is the issue.

PR #151, opened 2019-08-19
Disallow package-lock.json files on commit

arielshaqed:
See error message in `pre-commit`.  We disallow any package-lock.json
files, whether they're staged or not, as they indicate that your tree
is inconsistent with other developers and with CI.

In order to use this you must
```sh
git config --global --add core.hooksPath .githooks/
```


vogre:
Something like [overcommit](https://github.com/sds/overcommit) uses `export GIT_TEMPLATE_DIR="$(overcommit --template-dir)"`, which sounds cool.

arielshaqed:
> `git config --global --add core.hooksPath .githooks/` the issue with this as opposed to without `--global` is that it allows malicious repositories to run code on your machine, right?

This sounds bad enough that we'd want to prevent it.  So I'm rewriting the description to say *not* to run globally.

arielshaqed:
> > `git config --global --add core.hooksPath .githooks/` the issue with this as opposed to without `--global` is that it allows malicious repositories to run code on your machine, right?
> 
> This sounds bad enough that we'd want to prevent it. So I'm rewriting the description to say _not_ to run globally.

And also rewrote the actual commit.

Thanks; pulling!

PR #152, opened 2019-08-20
Test db client

arielshaqed:


local-proxy/src/test/flakyInvoke.test.ts
```diff
@@ -26,5 +26,5 @@ test('Regression #36', async (t) => {
     }))
   );
   t.is(responses.length, 100);
-  t.assert(responses.every((elem) => elem.body === 'hello'));
+//  t.assert(responses.every((elem) => elem.body === 'hello'));
```
vogre: Leftover commented out code.

local-proxy/src/server.ts
```diff
@@ -21,6 +23,14 @@ if (!localToken) {
   throw new Error('SHIFT_DEV_SERVER_LOCAL_TOKEN env var not defined');
 }
 
+function setupEnv({ port }: { port: number }) {
+  process.env.DB_ENDPOINT = `http://localhost:${port}/v1`;
```
vogre: `DB_ENDPOINT` -> `DB_BASE_URL`,  without `/v1`.

local-proxy/src/server.ts
```diff
@@ -21,6 +23,14 @@ if (!localToken) {
...
+  // TODO(vogre): Used (at least) for DB, where it doesn't really
+  //     matter locally.  Are better values available?
+  process.env.APP_ID = 'local-app';
+  process.env.ACCOUNT_ID = 'me';
+}
+
 const whitelistedModulesArr = (process.env.SHIFT_DEV_SERVER_MODULES_WHITELIST || '@binaris/shift-db').split(',');
```
vogre: Note that we should not load `@binaris/shift-db` before `setupEnv`, otherwise the environment variables are read during parsing and not on the first request.

local-proxy/src/server.ts
```diff
@@ -21,6 +23,14 @@ if (!localToken) {
   throw new Error('SHIFT_DEV_SERVER_LOCAL_TOKEN env var not defined');
 }
 
+function setupEnv({ port }: { port: number }) {
+  process.env.DB_ENDPOINT = `http://localhost:${port}/v1`;
+  // TODO(vogre): Used (at least) for DB, where it doesn't really
+  //     matter locally.  Are better values available?
+  process.env.APP_ID = 'local-app';
+  process.env.ACCOUNT_ID = 'me';
```
vogre: `ACCOUNT_ID` -> `API_KEY`.

local-proxy/src/server.ts
vogre: In fact, `@binaris/shift-db` should not be whitelisted at all it's `@binaris/shift-leveldb-server` which has the exposed `poll` that's required here.

interfaces/package.json
```diff
@@ -19,6 +19,7 @@
   "author": "",
   "license": "MIT",
   "dependencies": {
+    "concord": "~0.2.2",
```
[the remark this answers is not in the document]
vogre: Okay, apparently `npx` will run the version in `node_modules` if exists, and download and install latest if not exists.

The missing piece for it to properly work is that `generate_interfaces` script usually runs before `rush install`, so it will still use the latest since no package in node_modules.

db-client/package.json
```diff
@@ -22,12 +22,22 @@
   "author": "",
   "license": "MIT",
   "dependencies": {
-    "@binaris/shift-interfaces-node-client": "^0.4.0",
+    "@binaris/shift-interfaces-node-client": "^0.4.0-nudb.1",
```
vogre: No need for caret, for rush style dependencies.

local-proxy/package.json
```diff
@@ -17,7 +17,8 @@
     "@babel/cli": "^7.5.5",
     "@babel/core": "^7.5.5",
     "@babel/plugin-transform-modules-commonjs": "^7.5.0",
-    "@binaris/shift-db": "0.3.1",
+    "@binaris/shift-interfaces-koa-server": "^0.4.0-nudb.1",
+    "@binaris/shift-leveldb-server": "^0.0.0",
```
vogre: No need for caret for rush style dependencies.

db-client/package.json
```diff
@@ -8,7 +8,7 @@
   "scripts": {
```
vogre: Incorrect comment from previous PR - the `"version"` field of `db-client` should be the last published `0.3.1` and not `0.0.0` (so that rush will increment it, and not try to re-use already existing `0.0.1`.

local-proxy/package.json
```diff
@@ -17,7 +17,8 @@
     "@babel/cli": "^7.5.5",
     "@babel/core": "^7.5.5",
     "@babel/plugin-transform-modules-commonjs": "^7.5.0",
-    "@binaris/shift-db": "0.3.1",
```
vogre: `@binaris/shift-db` as direct dependency of `@binaris/shift-local-proxy` is really fragile and probably incorrect, but perhaps it can be listed in `peerDependencies` (since it still works correctly but might print a warning message `"Could not require whitelisted module @binaris/shift-db"`, WDYT?

db-client/package.json
arielshaqed: Good point.  Changed back and editing the change to be a minor bump (which is pretty major pre-1.0.0).

db-client/package.json
arielshaqed: Done!

interfaces/package.json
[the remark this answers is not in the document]
arielshaqed: It's a longer piece of spaghetti than that, though.  `generate_interfaces` bypasses rush there (again!) and `npm install`s.  https://github.com/binaris/shiftjs/blob/0aced8b806f09eb9044422bfe80ec3079450ebad/scripts/generate_interfaces#L4-L6
Yessss.  I am not proud of myself for this.

local-proxy/package.json
arielshaqed: Sounds reasonable enough.  Doing this.  I reckon eventually  we'll change the code only to emit warnings if a library is accessed.

common/changes/@binaris/shift-db-client/split-db_2019-08-18-08-47.json
```diff
@@ -3,7 +3,7 @@
     {
       "packageName": "@binaris/shift-db-client",
       "comment": "New client code for accessing over Concord",
-      "type": "patch"
+      "type": "minor"
```
vogre: This is the wrong `changes` file since we no longer have a `shift-db-client` package, but only shift-db. I think `common/changes/@binaris/shift-db-client` can be removed since it will have no corresponding package.

local-proxy/package.json
```diff
@@ -51,5 +52,8 @@
     "lodash": "^4.17.15",
     "tslint": "^5.18.0",
     "typescript": "^3.5.3"
+  },
+  "peerDependencies": {
+    "@binaris/shift-db": "~0.3.0 || ~0.4.0"
```
vogre: I think this should be `"0.3.1"` - it will be updated by `rush publish` automatically anyway to `"0.4.0"` during publishing.

db-client/package.json
```diff
@@ -22,12 +22,22 @@
...
-    "@binaris/shift-interfaces-node-client": "^0.4.0",
+    "@binaris/shift-interfaces-node-client": "0.4.0-nudb.1",
     "@types/deep-freeze": "^0.1.2",
     "deep-freeze": "0.0.1",
     "ramda": "^0.26.1"
   },
   "devDependencies": {
-    "@types/ramda": "^0.26.19"
+    "@binaris/shift-interfaces-koa-server": "^0.4.0-nudb.1",
+    "@binaris/shift-leveldb-server": "^0.0.0",
```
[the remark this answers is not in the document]
arielshaqed: Done (across the board).  Now `git grep '@binaris.*\^'` is clean

leveldb-server/package.json
```diff
@@ -15,7 +15,7 @@
   "author": "",
   "license": "MIT",
   "dependencies": {
-    "@binaris/shift-interfaces-koa-server": "^0.4.0",
+    "@binaris/shift-interfaces-koa-server": "^0.4.0-nudb.1",
```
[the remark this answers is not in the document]
arielshaqed: Gotcha.  "Intra-repo" :-)

local-proxy/package.json
arielshaqed: It seems to me that would be incorrect.  `peerDependencies` are supposed to specify supported versions -- they shouldn't move ahead just because a new version is available.  I hope Rush won't advance versions like this.

vogre: Rush updates peerDependencies, so a `rush publish -a` will make the change, but keep the `~`.
```
-    "@binaris/shift-db": "~0.3.0 || ~0.4.0"
+    "@binaris/shift-db": "~0.4.0"
```

This would allow us to release versions of `@binaris/shift-db` 0.4.1 0.4.2 0.4.3... When we'll make a minor change in `@binaris/shift-db` we'll probably need to release a new version of `@binaris/shift-local-proxy`.

I don't really want to support `~` in versions, I would like that it's possible to have exactly one combination of packages and not various combinations, but it probably needs to be solved on a different level.

arielshaqed: Not sure how that will work with vendoring.  Specifically, suppose I want to publish some ShiftJS-capable component.  I depend on some version of `shift-db`, but it will _not_ be the one that is necessarily installed in the app which uses my component.
Anyway, I want to defer fixing this to "later".

arielshaqed:
Thanks!
Pulling as soon as last round of tess ends.

PR #153, opened 2019-08-20
Include shift-db and dependencies in deploy 

vogre:
This is backwards compatible since all previously deployed apps already
had to avoid using shift-db

Needed as pre-requisite for deploying #152.


server-function/shift_deploy_binaris.sh
```diff
@@ -1,23 +0,0 @@
-#!/usr/bin/env bash
```
arielshaqed: (Just a question, I don't understand)

Why did we need this field, but not any more?

vogre: It was re-implemented in https://github.com/binaris/shiftjs/pull/138, so this is getting rid of the old implementation to avoid confusion.

PR #154, opened 2019-08-20
Fix log limits

michaeladda:
unfortunately only tested manually

### DIFF
--- a/cli/src/utils/cli-flags.ts
+++ b/cli/src/utils/cli-flags.ts
@@ -3,21 +3,23 @@ import { error } from '@oclif/errors';
 import { isISO8601, isInt } from 'validator';
 import ms from 'ms';
 
-export type MinMaxIntFlagOptions = Parameters<(typeof flags.integer)>[0] & {
+export type MinMaxIntFlagOptions = Parameters<(typeof flags.integer)>[0] & Partial<{
   min: number,
   max: number,
-};
+}>;
 
 export default {
   ...flags,
-  minMaxInt: ({ min, max }: MinMaxIntFlagOptions) => flags.integer({
+  minMaxInt: (options: MinMaxIntFlagOptions) => flags.build({
+    ...options,
     parse(val) {
+      const { min, max } = options;
       if (!isInt(val, { min, max })) {
         error(`Expected an integer between ${min} and ${max} but received: ${val}`);
       }
       return parseInt(val, 10);
     },
-  }),
+  })(),
   durationOrISO8601: flags.build({
     parse(val) {
       if (isISO8601(val)) {
--- a/cli/src/utils/cli-flags.ts
+++ b/cli/src/utils/cli-flags.ts
@@ -12,7 +12,6 @@ export default {
   ...flags,
   // Using flags.build since flags.integer does not honor the default
   minMaxInt: (options: MinMaxIntFlagOptions) => flags.build({
-    ...options,
     parse(val) {
       const { min, max } = options;
       if (!isInt(val, { min, max })) {
@@ -20,6 +19,7 @@ export default {
       }
       return parseInt(val, 10);
     },
+    ...options,
   })(),
   durationOrISO8601: flags.build({
     parse(val) {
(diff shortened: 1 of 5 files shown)


cli/src/utils/cli-flags.ts
```diff
@@ -3,21 +3,23 @@ import { error } from '@oclif/errors';
...
+    ...options,
     parse(val) {
+      const { min, max } = options;
       if (!isInt(val, { min, max })) {
         error(`Expected an integer between ${min} and ${max} but received: ${val}`);
       }
       return parseInt(val, 10);
     },
-  }),
+  })(),
```
arielshaqed: I don't understand this.  It looks like you're calling the return value of `flags.build` as a function with no args, and using the return of _that_ function as the value of `minMaxInt`.  How does this even work (or compile, assuming reasonable types)?

Or, why do you need to call the return value from `flags.build` here, but not below in `durationOrISO8601`?

arielshaqed: Yuck.  Thanks for explaining!

cli/src/utils/cli-flags.ts
```diff
@@ -3,21 +3,23 @@ import { error } from '@oclif/errors';
...
+export type MinMaxIntFlagOptions = Parameters<(typeof flags.integer)>[0] & Partial<{
   min: number,
   max: number,
-};
+}>;
 
 export default {
   ...flags,
-  minMaxInt: ({ min, max }: MinMaxIntFlagOptions) => flags.integer({
+  minMaxInt: (options: MinMaxIntFlagOptions) => flags.build({
```
arikmaor: why change this?

cli/src/utils/cli-flags.ts
```diff
@@ -3,21 +3,23 @@ import { error } from '@oclif/errors';
 import { isISO8601, isInt } from 'validator';
 import ms from 'ms';
 
-export type MinMaxIntFlagOptions = Parameters<(typeof flags.integer)>[0] & {
+export type MinMaxIntFlagOptions = Parameters<(typeof flags.integer)>[0] & Partial<{
```
arikmaor: If your'e not calling `flags.integer` I guess `Parameters<(typeof flags.integer)>[0]` should be changed also.
BUT, I don't understand why you're changing the code here at all...

cli/src/utils/cli-flags.ts
```diff
@@ -3,21 +3,23 @@ import { error } from '@oclif/errors';
...
-};
+}>;
 
 export default {
   ...flags,
-  minMaxInt: ({ min, max }: MinMaxIntFlagOptions) => flags.integer({
+  minMaxInt: (options: MinMaxIntFlagOptions) => flags.build({
+    ...options,
     parse(val) {
+      const { min, max } = options;
```
arikmaor:
```suggestion
```

cli/src/utils/cli-flags.ts
arikmaor:
```suggestion
  minMaxInt: ({ min, max, ...options }: MinMaxIntFlagOptions) => flags.integer({
```

cli/src/utils/cli-flags.ts
```diff
@@ -3,21 +3,23 @@ import { error } from '@oclif/errors';
...
   min: number,
   max: number,
-};
+}>;
 
 export default {
   ...flags,
-  minMaxInt: ({ min, max }: MinMaxIntFlagOptions) => flags.integer({
+  minMaxInt: (options: MinMaxIntFlagOptions) => flags.build({
+    ...options,
```
arikmaor: consider moving after `parse`

cli/src/utils/cli-flags.ts
arikmaor:
```suggestion
  }),
```

cli/src/utils/cli-flags.ts
michaeladda: the default is an important option, and without using build the default does not take place

cli/src/utils/cli-flags.ts
michaeladda: sorry, won't fly. build needs an immediate call (integer doesn't of course). See my reply above.

PR #158, opened 2019-08-21
Tell Linguist (and GitHub) that shrinkwrap.json is generated

arielshaqed:
This lets GitHub hide diffs in it by default.

Also my teleprinter is broken: get rid of CRs.

It looks like this (I pushed a bad change to shrinkwrap and now I'm resetting it, so you can see for yourself on this very PR):
image


arielshaqed:
Thanks!

PR #159, opened 2019-08-21
Update cli README after publish

vogre:
Manual until integrated into publish process


vogre:
I didn't run `rush change` intentionally since this isn't supposed to generate a changelog (even an empty one). After merge to `master` there will be no warning of changes, since we are always comparing against `master`.

PR #160, opened 2019-08-21
Use upstream nodemon

vogre:
There are no new actual upstream changes
Use a more precise version number
Prevent various npm bugs installing npm packages


vogre:
Previously the patched github version allowed to get to the `ChildProcess` object, currently this functionality is unused.

PR #164, opened 2019-08-23
Upload less files

vogre:
Avoid a flattened copy of build folder
Ignore all .dotfiles
Avoid adding bundle.tgz to bundle.tgz

### DIFF
--- a/cli/src/commands/deploy.ts
+++ b/cli/src/commands/deploy.ts
@@ -47,7 +47,8 @@ export default class Deploy extends Command {
       const filesToExclude = new Set(['node_modules', 'backend', 'src'].map((f) => pathResolve(projectDir, f)));
       await copy(projectDir, stagingDir, {
         filter(src) {
-          return !filesToExclude.has(src) && !path.basename(src).startsWith('.');
+          return !filesToExclude.has(src) &&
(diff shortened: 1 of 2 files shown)


cli/src/commands/deploy.ts
```diff
@@ -46,11 +45,9 @@ export default class Deploy extends Command {
       }
 
       const filesToExclude = new Set(['node_modules', 'backend', 'src'].map((f) => pathResolve(projectDir, f)));
-      const dotShiftPath = pathResolve(projectDir, '.shift');
       await copy(projectDir, stagingDir, {
         filter(src) {
-          // TODO: we should just put all the .shift files in a single .shift directory
-          return !filesToExclude.has(src) && !src.startsWith(dotShiftPath);
+          return !filesToExclude.has(src) && !path.basename(src).startsWith('.');
```
michaeladda: won't this also skip user files starting with dot under `build` as well?
maybe skip only dotfiles in the topmost directory? (for .git and shift related stuff)

vogre: Thanks!

There are really quite subtle behaviors that need to keep track of when translating from bash.

PR #166, opened 2019-08-26
Clean cli min max flags

michaeladda:
as requested in #154 
manually tested similar to 154


michaeladda:
similar full manual test as in 154 and as written above :-p

PR #167, opened 2019-08-26
Don't copy Operation definition from json-fast-patch

bergundy:


arielshaqed:
Also make sure that `Operation`s are in fact generated.

PR #169, opened 2019-08-26
Rename DB and Handler, restore query

bergundy:


db-client/src/index.ts
```diff
@@ -1,110 +1,29 @@
...
+// TODO(bergundy): Verify environment variables
+const db = new DB(
+  `${process.env.DB_BASE_URL!}/v1`,
+  {
     auth: {
       v1: {
-        // TODO(ariels): Update with prefixes from binaris/spice#616.
-        appId: process.env.APP_ID!,
-        apiKey: process.env.API_KEY!,
+        appId: process.env.SHIFT_APP_ID!,
```
arielshaqed: This has already changed in #168 (which follows Lycan), we should follow that one

PR #170, opened 2019-08-27
Change auth field of ClientContext in DB interface

bergundy:


interfaces/src/db.ts
```diff
@@ -105,11 +105,10 @@ export interface Query {
 
 // tslint:disable-next-line:no-empty-interface
 export interface ClientContext {
+  appId: string;
   auth: {
-    v1: {
-      // Contents TBD
-      appId: string;
-      apiKey: string;
+    v1?: {
```
arielshaqed: If this is optional then you needn't include it everywhere.

interfaces/src/db.ts
```diff
@@ -105,11 +105,10 @@ export interface Query {
 
 // tslint:disable-next-line:no-empty-interface
 export interface ClientContext {
+  appId: string;
```
arielshaqed: Occurs to me: This might be better inside another object: it's the ID for what you want to access.  E.g. that subobject may expand to include collection ID.

interfaces/src/db.ts
bergundy: It's optional so you can use `v2` instead of `v1` once we specify it.

interfaces/src/db.ts
bergundy: We could just add `collection` to the top level when we support it.

PR #171, opened 2019-08-27
Add env and collection to DB context, fix names of env vars in local-proxy

bergundy:


db-client/src/test/component/db.test.ts
```diff
@@ -43,6 +43,8 @@ test.beforeEach(async (t) => {
     `http://localhost:${port}/v1`,
     {
       appId: 'testing',
+      appEnv: 'default',
+      collection: 'default',
```
vogre: Might be confusing that both env and collection are default...

db-client/src/index.ts
```diff
@@ -9,14 +9,23 @@ import { DB, Versioned, Q } from './db';
 
 export { Q };
 
-// TODO(bergundy): Verify environment variables
+function assertEnv(name: string): string {
```
vogre: Better than undefineds, although still missing a test of shift-db in local-proxy (not necessarily for this PR).

bergundy: #172

db-client/src/test/component/db.test.ts
bergundy: Hmm.. that's what we use ATM in the deployed version.

PR #174, opened 2019-08-27
Share DB testsuite

bergundy:


db-testsuite/src/index.ts
```diff
@@ -27,41 +21,52 @@ async function listenOn(app: Koa): Promise<Server> {
...
+    const { context, handler } = await setUp();
+    const dbRouter = new DBRouter(handler, true);
+    const router = new KoaRouter();
+    router.use('/v1', dbRouter.koaRouter.routes(), dbRouter.koaRouter.allowedMethods());
+    const app = new Koa();
+    app.use(router.routes());
+    app.use(router.allowedMethods());
+
+    const server = await listenOn(app);
+    const port = (server.address() as unknown as AddressInfo).port;
```
vogre: Redundant `as unknown`?

db-testsuite/src/index.ts
```diff
@@ -1,23 +1,17 @@
...
+import { DB } from '@binaris/shift-db/dist/db';
 
 interface Context {
-  dbDir: string;
   client: DB;
+  injectedContext: any;
 }
 
-const test = anyTest as TestInterface<Context>;
+export const test = anyTest as TestInterface<Context>;
```
vogre: Extra `export`?

bergundy: Nope, you can add `test.before` and other hooks this way.

db-testsuite/src/index.ts
bergundy: Maybe, I copied this as is.

db-client/src/db.ts
```diff
@@ -69,7 +69,7 @@ export class DB {
       // it by referring to value in an object.
       const newValue = updater(deepFreeze({ value: value as T }).value);
       checkValue(newValue);
-      if (await this.setIfVersion(key, newValue, version)) return deepFreeze(newValue);
+      if (await this.setIfVersion(key, version, newValue)) return deepFreeze(newValue);
```
vogre: [Sonarts](https://github.com/SonarSource/SonarTS/blob/HEAD/sonarts-core/docs/rules/arguments-order.md) rule might prevent this. If type checking is not enough perhaps worth making them named instead of positional arguments.

bergundy: I changed the signature to look like the one in `interfaces/src/db.ts`.

db-testsuite/src/index.ts
bergundy: Used by thoth.

PR #175, opened 2019-08-28
Restore local server polling capabilities

bergundy:


db-client/src/db.ts
```diff
@@ -81,19 +83,40 @@ export class DB {
...
+  /**
+   * Polls on updates to specified keys since specified versions.
+   * @see KeyedVersions
+   * @see KeyedPatches
+   */
+  public async poll(
+    keysToVersions: Array<[string, Version]>,
+    opts: PollOptions = {},
+  ): Promise<Array<[string, Patch[]]>> {
+    return this.client.poll(this.ctx, keysToVersions, opts);
```
arielshaqed: Nit: `async` _and_ not `await`?

db-testsuite/src/index.ts
```diff
@@ -185,7 +203,120 @@ test('DB.update does nothing if document not updated', async (t) => {
...
+  const { client } = t.context;
+  await client.create('test1', 'a');
+  const { version } = (await client.startPolling('test1'))!;
+  const [keyedPatches] = await Promise.all([
+    client.poll([['test1', version]]),
+    client.remove('test1'),
+  ]);
+  t.deepEqual(keyedPatches, [
+    ['test1', [
+      { version: { major: version.major, minor: 2 }, ops: [{ op: 'remove', path: '/root' }] },
```
arielshaqed: Isn't this `incrVersion(version)`?

db-testsuite/src/index.ts
```diff
@@ -219,3 +350,13 @@ test('DB.update throws TypeError if trying to modify returned object', async (t)
     return obj;
   }), TypeError);
 });
+
+test('DB.update sets operartionId', requiresPolling, async (t) => {
```
arielshaqed:
```suggestion
test('DB.update sets operationId', requiresPolling, async (t) => {
```

db-client/src/db.ts
bergundy: I agree with @vogre that this way is cleaner.

db-testsuite/src/index.ts
bergundy: Yes, thanks

db-testsuite/src/index.ts
bergundy: Thanks!

PR #176, opened 2019-08-28
Add no-return-await rule

vogre:
Perf: do not create extra promises
Ensure code is consistent (without the rule both variants are valid)

### DIFF
--- a/common/changes/@binaris/shift-db/no-return-await_2019-08-28-09-21.json
+++ b/common/changes/@binaris/shift-db/no-return-await_2019-08-28-09-21.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "comment": "",
+      "packageName": "@binaris/shift-db",
+      "type": "none"
+    }
+  ],
+  "packageName": "@binaris/shift-db",
+  "email": "vladimir@shiftjs.com"
+}
--- a/common/changes/@binaris/shift-leveldb-server/no-return-await_2019-08-29-08-09.json
+++ b/common/changes/@binaris/shift-leveldb-server/no-return-await_2019-08-29-08-09.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "comment": "",
+      "packageName": "@binaris/shift-leveldb-server",
+      "type": "none"
+    }
+  ],
+  "packageName": "@binaris/shift-leveldb-server",
+  "email": "vladimir@shiftjs.com"
+}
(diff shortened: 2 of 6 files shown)


db/src/db.ts
```diff
@@ -145,7 +145,7 @@ export class DB extends EventEmitter {
    */
   public async create(key: string, value: Serializable): Promise<boolean> {
```
arielshaqed: This function no longer needs to be `async`.  What do you (we) think about that?  (You mentioned perf, so AFAIU this _does_ create an extra `Promise`.  I don't know if the JITs don't strip them out, obviously.)
```
> async function af() { return 17; }
undefined
> af()
Promise {
  17,
  domain:
   Domain {
     domain: null,
     _events:
      [Object: null prototype] {
        removeListener: [Function: updateExceptionCapture],
        newListener: [Function: updateExceptionCapture],
        error: [Function: debugDomainError] },
     _eventsCount: 3,
     _maxListeners: undefined,
     members: [],
     [Symbol(kWeak)]: WeakReference {} } }
```

bergundy:
Does this rule apply to?
```js
try {
  return await promise;
} catch (err) {
  // error occurred asynchronously
}

**Comment by vogre** [comment]
No, the rule is smart enough to allow `try-await-catch`.

File: db/src/db.ts (hunk 1 above)
Comment (arielshaqed) [reply db/src/db.ts]: > There is a subtle difference in unmaking it `async` - if `checkValue` throws it's going to be an exception rather than a rejection. Usually it doesn't matter that much, but should not be changed without checking.

I agree that there's a difference, and actually I think it's an improvement (thanks!).  I don't think that we depend on the existing behaviour.  And I *do* believe that this code is (potentially) useful -- making the "sync promise" style the most expressive for something like this:
```ts
function setVals(vs: any[]) {
  try {
    return vs.map((val, idx) => create(`k${idx}`, val));
  } catch (e) {
    if (...) throw new TypeError('some value was bad');
  }
}
function waitAndSet(vs: any[]) {
  try {
    await Promise.all(setVals());
    return true;
  } catch (e) {
    if (e instanceof TypeError) throw e;
    log('bad db', e);
    return false
  }
}
```

If you agree that this style is best, can we just specify it *now*, so that we never have to change it later?

vogre:
Thanks!

AFAICT the only 100% safe case where `async` keyword can be removed is a `return new Promise(...)`, so I don't think we should remove `async` from functions that don't contain `await`.

Currently merging like this. Will open a separate PR to consider [no-async-without-await](https://palantir.github.io/tslint/rules/no-async-without-await/).

PR #177, opened 2019-08-28
Fix and test db-client find()

bergundy:


interfaces/src/db.ts
```diff
@@ -1,5 +1,5 @@
 import { Operation } from 'fast-json-patch';
-export type Comparable = string | number | Date;
+export type Comparable = string | number;
```
vogre: The fix is removal of Date from types?

db-testsuite/src/index.ts
```diff
@@ -7,7 +7,7 @@ import { range } from 'ramda';
 import nanoid from 'nanoid';
 import { DBRouter, DBHandler } from '@binaris/shift-interfaces-koa-server';
 import { Version } from '@binaris/shift-interfaces-koa-server/interfaces';
-import { DB } from '@binaris/shift-db/dist/db';
+import { DB, Q } from '@binaris/shift-db/dist/db';
```
vogre: Possible to avoid exposing the implementation having `dist/` by adding a `dist.js` which does `module.exports = require('./dist/db.js');` - not sure whether really needed.

interfaces/src/db.ts
bergundy: Yes... it was never officially supported and will have to be revisited.

db-testsuite/src/index.ts
bergundy: I think it's safe since it's in the same repo.

PR #180, opened 2019-08-29
Share DB match tests

bergundy:


leveldb-server/src/test/client.test.ts
```diff
@@ -4,7 +4,7 @@ import { tmpdir } from 'os';
 import { promisify } from 'util';
 import * as path from 'path';
 import { Handler } from '../db';
-import { setUpTests } from '@binaris/shift-db-testsuite';
+import { setUpTests } from '@binaris/shift-db-testsuite/dist/client';
```
michaeladda: you'll need to do the same in thoth (didn't see a matching PR)

db-testsuite/src/match.ts
```diff
@@ -0,0 +1,248 @@
...
+    t.false(await t.context.match({ key: 'abc', value: {} }, filter));
+  }
+  for (const path of [['key'], ['value']]) {
+    const filter = { operator: 'exists' as 'exists', path };
+    t.true(await t.context.match({ key: 'abc', value: {} }, filter));
+  }
+});
+
+test('isNull checks for null', predicateHolds, null, 'isNull');
+test('isNull checks for null', predicateFails, 0, 'isNull');
```
michaeladda: somehow `test('isNull checks for null', predicateFails, undefined, 'isNull');` got dropped here during the move

bergundy: Changed the tests up, you can't create a document with an `undefined` value.

PR #181, opened 2019-08-29
Add list, browse and whoami CLI commands

michaeladda:
unfortunately only tested manually.
`./bin/run list a`
`./bin/run list --a`
`./bin/run list`
`./bin/run whoami`
`./bin/run browse`
also with fake templates data and extra flag/arg to whoami and browse commands

### DIFF
--- a/cli/src/commands/browse.ts
+++ b/cli/src/commands/browse.ts
@@ -1,0 +1,28 @@
+import Command from '../utils/command';
+import terminalLink from 'terminal-link';
+
+export default class Browse extends Command {
+  public static description = 'list templates';
+  public static hidden = false;
+  public static examples = [
+    `$ ${Command.cliBinName} browse`,
+  ];
+
+  public static args = [];
+
+  public static strict = true;
+
+  public async run() {
+    this.parse(Browse);
+    const templates = await this.lycanClient.listTemplates();
+    if (templates.length === 0) {
+      this.log('Sorry. Templates are not available yet.');
+      return;
+    }
+    for (const { id, name, previewImageUrl, githubUrl} of templates) {
+      const previewImageLink = terminalLink('preview image', previewImageUrl);
+      const githubLink = terminalLink('source', githubUrl);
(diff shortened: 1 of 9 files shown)


cli/src/commands/browse.ts
```diff
@@ -0,0 +1,28 @@
+import Command from '../utils/command';
+import terminalLink from 'terminal-link';
+
+export default class Browse extends Command {
+  public static description = 'list templates';
+  public static hidden = false;
+  public static examples = [
```
vogre: Redundant, as the output for EXAMPLES becomes the same as USAGE? (similar for other `examples` for commands with no arguments)

cli/src/commands/browse.ts
```diff
@@ -0,0 +1,28 @@
...
+    this.parse(Browse);
+    const templates = await this.lycanClient.listTemplates();
+    if (templates.length === 0) {
+      this.log('Sorry. Templates are not available yet.');
+      return;
+    }
+    for (const { id, name, previewImageUrl, githubUrl} of templates) {
+      const previewImageLink = terminalLink('preview image', previewImageUrl);
+      const githubLink = terminalLink('source', githubUrl);
+      this.log(`${name}: See ${previewImageLink}, get the ${githubLink} from GitHub, or try it with '${Command.cliBinName} try ${id}'`);
```
vogre: `try` command will be in a future PR?

Also, maybe requires highlighting of a shell command like this: `$ ${Command.cliBinName}`

cli/src/commands/browse.ts
michaeladda: keeping it for now to be consistent with other commands. once we decide on examples in general we can wipe everything out.

cli/src/commands/browse.ts
michaeladda: indeed, try is in the works.
the highlighting is with ticks (i.e. `'foo'`) for easy of copy paste. leaving as is.

PR #182, opened 2019-09-01
Disable deep-freezing of update() params

### LINKED ISSUE #162: Incoming arg to updater cannot be modified
When using the update function from ShiftDB, the `prevState` argument appears to be 'read-only'. I'm unaware of any spec which defines such functionality, so as of now I assume this is unintended.

```javascript
update((someData = { key: [] }) => {
  someData.key.push('item');
  return someData;
});
```

The above code results in the following error

```javascript
TypeError: Cannot assign to read only property 'key' of #<Object>
```

I suggest initially and immediately reverting this to the specced behavior, and then we can open a discussion regarding future design.

bergundy:
Closes #162

### DIFF
--- a/db-client/src/index.ts
+++ b/db-client/src/index.ts
@@ -1,5 +1,4 @@
 import process from 'process';
-import { DeepReadonly } from 'deep-freeze';
 import {
   Document,
   Patch,
@@ -68,8 +67,8 @@ export async function remove(key: string): Promise<boolean> {
  * @return - The new value returned from updater
  */
 export async function update<T extends Serializable = any>(
-  key: string, updater: (state?: DeepReadonly<T>) => T, options?: UpdateOptions,
-): Promise<DeepReadonly<T>> {
+  key: string, updater: (state?: Readonly<T>) => T, options?: UpdateOptions,
+): Promise<Readonly<T>> {
   return db.update(key, updater, options);
 }
 // Available only on backend, needs to pass a function.
--- a/common/changes/@binaris/shift-db-testsuite/no-deep-freeze_2019-09-01-10-22.json
+++ b/common/changes/@binaris/shift-db-testsuite/no-deep-freeze_2019-09-01-10-22.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@binaris/shift-db-testsuite",
+      "comment": "Remove deep-freeze test",
+      "type": "patch"
+    }
+  ],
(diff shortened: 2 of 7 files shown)


db-client/src/index.ts
```diff
@@ -68,8 +67,8 @@ export async function remove(key: string): Promise<boolean> {
  * @return - The new value returned from updater
  */
 export async function update<T extends Serializable = any>(
-  key: string, updater: (state?: DeepReadonly<T>) => T, options?: UpdateOptions,
-): Promise<DeepReadonly<T>> {
+  key: string, updater: (state?: Readonly<T>) => T, options?: UpdateOptions,
```
arielshaqed: I'm not sure we can do even this much.
1. It makes code allowed in JavaScript not allowed in TypeScript (you need to cast away `readonly`ness).
2. Will it work for JavaScript on TypeScript-aware editors?  If they _complain_ that's bad enough (we've decided to allow this behaviour AFAIU), but if they _stop being type-aware_ then it's really bad.
3. We wanted to use TypeScript to validate JSDoc types in our JavaScript templates by running the TypeScript compiler on them.  Will that work?

Basically, I'm not sure what we agreed to do with @ashevat, but I'm pretty sure this change is a bad result.  You could go all the way and remove `Readonly` entirely.

vogre: I think the function is quite clunky, since you always need to specify the <T> type parameter. I suggest we postpone the discussion of TypeScript ergonomics to some point in the future to allow this to go forward.

> It makes code allowed in JavaScript not allowed in TypeScript (you need to cast away readonlyness).

If the behavior is something we don't recommend, it can be thought of more like a TypeScript feature - it prevents you from doing something dangerous (of course when you need to progressively transition to TypeScript you can start with casting `as any` all the complaints and fix them at some point later).

> Will it work for JavaScript on TypeScript-aware editors? If they complain that's bad enough (we've decided to allow this behaviour AFAIU), but if they stop being type-aware then it's really bad.

VSCode won't complain about mismatching types. As it currently goes the compiler cannot infer <T>, so JavaScript code sees a `Readonly<any>` instead of `DeepReadonly<any>`.

> We wanted to use TypeScript to validate JSDoc types in our JavaScript templates by running the TypeScript compiler on them. Will that work?

Compiling JS templates as TypeScript would prohibit mutations, which is probably what we recommend. The state is still inferred as `Readonly<any>`, so no special typechecking happens beyond that.

PR #184, opened 2019-09-01
Add try and claim CLI commands

michaeladda:
manually tested with matching binaris/spice#644

### DIFF
--- a/cli/src/commands/claim.ts
+++ b/cli/src/commands/claim.ts
@@ -1,0 +1,28 @@
+import Command from '../utils/command';
+import terminalLink from 'terminal-link';
+
+export default class Claim extends Command {
+  public static description = 'claim an application';
+  public static hidden = false;
+  public static examples = [
+    `$ ${Command.cliBinName} claim OiJIUzI1NiIsIn`,
+  ];
+
+  public static args = [
+    {
+      name: 'token',
+      required: true,
+      description: 'claim token',
+    },
+  ];
+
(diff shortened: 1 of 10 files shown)


cli/src/commands/list.ts
```diff
@@ -20,8 +20,8 @@ export default class List extends Command {
       return;
     }
     // TODO: Get URL from lycan
```
bergundy: Remove the TODO please.

cli/src/commands/claim.ts
```diff
@@ -0,0 +1,28 @@
...
+      name: 'token',
+      required: true,
+      description: 'claim token',
+    },
+  ];
+
+  public static strict = true;
+
+  public async run() {
+    const { args } = this.parse(Claim);
```
bergundy: You should check that you're authenticated here: https://github.com/binaris/shiftjs/blob/b76a1672cde2422afa48e346043584ab9f766ac9/cli/src/commands/deploy.ts#L116

cli/src/commands/list.ts
bergundy: You should authenticate() above as well.

michaeladda:
Thanks for review. Addressed issues and found a few additional similar problems so fixed those too. @bergundy PTAL :-)

michaeladda:
Excellent. Waiting for binaris/spice#644 before merging

PR #185, opened 2019-09-02
Post publish changes

michaeladda:


common/changes/shift-cli/master_2019-08-29-12-38.json
```diff
@@ -1,11 +0,0 @@
-{
-  "changes": [
```
vogre: This is odd, it should have been included in the previous commit: https://github.com/binaris/shiftjs/commit/0cfeb74cf366d1aca113fade50c4905ffefd0f33.

PR #186, opened 2019-09-02
Deploy non-JS files of backend directory

vogre:
Also allow access to non-JS files in backend directory from local-proxy

Discovered by @rylandg

### DIFF
--- a/cli/src/commands/deploy.ts
+++ b/cli/src/commands/deploy.ts
@@ -72,6 +72,12 @@ export default class Deploy extends Command {
         shell: process.platform === 'win32',
       });
 
+      await copy(pathResolve(projectDir, 'backend'), pathResolve(stagingDir, 'backend'), {
+        filter(src) {
+          return path.extname(src) !== '.js';
+        },
+      });
+
       return stagingDir;
 
     } catch (err) {
--- a/local-proxy/package.json
+++ b/local-proxy/package.json
@@ -24,6 +24,7 @@
     "address": "^1.1.2",
     "bunyan": "^1.8.12",
     "bunyan-rotating-file-stream": "^1.6.3",
+    "env-var": "^5.0.0",
     "express": "^4.17.1",
     "fs-extra": "^8.1.0",
     "http-proxy": "^1.17.0",
--- a/local-proxy/package.json
+++ b/local-proxy/package.json
@@ -25,6 +25,7 @@
     "bunyan": "^1.8.12",
     "bunyan-rotating-file-stream": "^1.6.3",
     "express": "^4.17.1",
+    "fs-extra": "^8.1.0",
     "http-proxy": "^1.17.0",
     "koa": "^2.7.0",
     "koa-bodyparser": "^4.2.1",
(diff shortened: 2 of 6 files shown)


local-proxy/package.json
```diff
@@ -25,6 +25,7 @@
     "bunyan": "^1.8.12",
     "bunyan-rotating-file-stream": "^1.6.3",
     "express": "^4.17.1",
+    "fs-extra": "^8.1.0",
```
rylandg: I know that @arielshaqed has strong feelings about this package. Is there another alternative?

local-proxy/src/server.ts
```diff
@@ -13,8 +13,9 @@ import os from 'os';
 import { initRegistry } from './stdio';
 import nanoid from 'nanoid';
 import mkdirp from 'mkdirp';
+import { copy } from 'fs-extra';
 
-const basePath = process.env.SHIFT_DEV_SERVER_BASE_REQUIRE_PATH;
+const basePath = process.env.SHIFT_DEV_SERVER_BASE_REQUIRE_PATH as string;
```
rylandg: Bit weird to cast as string but then do a truthy check on it. Why not cast after checking?

rylandg: Also there should be a generic way to do this.

local-proxy/src/server.ts
```diff
@@ -72,17 +73,25 @@ const router = new KoaRouter();
...
+    },
+  });
+  await copy(basePath, tmpDir, {
+    filter(src) {
+      return extname(src) !== '.js';
+    },
+  });
+}
 
+const transpilePromise = transpileAndCopy();
```
rylandg: Not 100% sure, but is this code as safe as it was before? Previously there was only a single promise, so if it fails it can manage its own cleanup. What happens if the `babelDir` succeeds but then the `copy` fails?

cli/src/commands/deploy.ts
```diff
@@ -73,6 +73,12 @@ export default class Deploy extends Command {
         shell: process.platform === 'win32',
       });
 
+      await copy(pathResolve(projectDir, 'backend'), pathResolve(stagingDir, 'backend'), {
```
rylandg: Frustrating that this has to essentially be duplicated. Not really solvable short term unfortunately.

local-proxy/package.json
vogre: There's `ncp`, but I don't see a reason to change since we are already using this one.

cli/src/commands/deploy.ts
vogre: There are subtle differences in the code around it so deduplication won't help that much here...

rylandg: That's why I said frustrating. At the end of the day, the code is very similar, minus subtle differences. Thats where the "Frustrating" came from.

local-proxy/package.json
rylandg: Makes sense

local-proxy/src/server.ts
rylandg: Cool

PR #189, opened 2019-09-04
Initial CLI tests (usage)

arielshaqed:
Need (at least) `remove` command to be able to test `deploy`, `logs`.

### DIFF
--- a/cli/src/test/cli.spec.ts
+++ b/cli/src/test/cli.spec.ts
@@ -4,13 +4,14 @@ import * as R from 'ramda';
 
 const process = R.evolve({ out: (x: Buffer) => x.toString(), err: (x: Buffer) => x.toString() });
 
-test('cli with no args shows VERSION, USAGE, COMMANDS', async (t) => {
+test('cli with no args shows help', async (t) => {
   const shell = new Shell();
   // TODO(ariels): Find this directory!
   const { out, err, ...status } = process(await shell.run('./bin/run'));
 
   t.true(success(status));
   t.assert(err === '');
+  t.assert(out.match(/^ShiftJS CLI Tool.*?VERSION.*?USAGE.*?COMMANDS/s));
 });
 
 test('cli with no args and cli help and cli --help give same output', async (t) => {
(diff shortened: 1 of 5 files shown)


cli/src/test/cli.spec.ts
```diff
@@ -0,0 +1,40 @@
+import test from 'ava';
+import { Shell, success } from 'specshell';
+import * as R from 'ramda';
+
+const process = R.evolve({ out: (x: Buffer) => x.toString(), err: (x: Buffer) => x.toString() });
+
+test('cli with no args shows VERSION, USAGE, COMMANDS', async (t) => {
```
bergundy: That's not what this tests

cli/src/test/cli.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+  const { out, err, ...status } = process(await shell.run('./bin/run'));
+
+  t.true(success(status));
+  t.assert(err === '');
+});
+
+test('cli with no args and cli --help give same output', async (t) => {
+  const shell = new Shell();
+  const noArgs = process(await shell.run('./bin/run'));
+  const helpArg = process(await shell.run('./bin/run'));
```
bergundy: Where is the `--help` arg?

cli/src/test/cli.spec.ts
arielshaqed: Sorry, pushed the pre-rebase version.

cli/src/test/cli.spec.ts
arielshaqed: Added *both* of them.

cli/src/test/cli.spec.ts
bergundy: Still wrong..
```suggestion
test('cli with no args exits successfully', async (t) => {
```

arielshaqed: Yeah, that test dropped an early assertion.  Restored that one.

arielshaqed:
Thanks!

PR #197, opened 2019-09-08
Add user-agent to cli api requests

vogre:


cli/src/utils/command.ts
```diff
@@ -47,9 +47,13 @@ export default abstract class BaseCommand extends Command {
 
   public get apiHeaders(): Record<string, string> {
     const apiKey = this.conf.get('accessToken') as string | undefined;
-    return apiKey !== undefined ? {
-      'shift-api-key': apiKey,
-    } : {};
+    const headers: Record<string, string> = {
+      'user-agent': `${pjson.name}/${pjson.version}`,
+    };
+    if (apiKey) {
```
arikmaor: it's gonna be nicer if this `if` was right after `const apiKey = ...`

PR #198, opened 2019-09-08
Don't use unexported type BabelTypes.BaseNode

arielshaqed:
Instead:

1. In one instance _explicitly_ type out the desired `leadingComments`
   field (to identify comments with `@expose`),
2. In the other instance use `object`: `isExpressionStatement` anyway
   expects that type and will cast to an `ExpressionStatement`.

### DIFF
--- a/code-transform/src/common.ts
+++ b/code-transform/src/common.ts
@@ -4,11 +4,15 @@ export function getFunctionName(e: BabelTypes.FunctionDeclaration, t: typeof Bab
   return t.isIdentifier(e.id) ? e.id.name : undefined;
 }
 
-export function isExposedStatement(node: BabelTypes.BaseNode) {
+interface WithBabelLeadingComments {
+  leadingComments: ReadonlyArray<BabelTypes.Comment> | null;
+}
+
+export function isExposedStatement(node: WithBabelLeadingComments) {
   return node.leadingComments && node.leadingComments.some((comment) => /@expose/.test(comment.value));
 }
 
-export function isTypeScriptGeneratedExport(e: BabelTypes.BaseNode, t: typeof BabelTypes, funcName: string) {
+export function isTypeScriptGeneratedExport(e: object, t: typeof BabelTypes, funcName: string) {
   return t.isExpressionStatement(e) &&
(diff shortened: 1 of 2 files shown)


code-transform/src/common.ts
```diff
@@ -4,15 +4,11 @@ export function getFunctionName(e: BabelTypes.FunctionDeclaration, t: typeof Bab
...
-  leadingComments: ReadonlyArray<BabelTypes.Comment> | null;
-}
-
-export function isExposedStatement(node: WithBabelLeadingComments) {
+export function isExposedStatement(node: BabelTypes.Statement) {
   return node.leadingComments && node.leadingComments.some((comment) => /@expose/.test(comment.value));
 }
 
-export function isTypeScriptGeneratedExport(e: object, t: typeof BabelTypes, funcName: string) {
+export function isTypeScriptGeneratedExport(e: BabelTypes.Statement, t: typeof BabelTypes, funcName: string) {
```
vogre: Actually used in `index.ts` as `sibling.node` where it's a `BabelTypes.Node`. (TBH That's a very fragile typescript output check 😞 )

arielshaqed:
> Thanks!
> 
> I think that we can in both cases use `BabelTypes.Statement` - it is more narrow than `Node` but the widest type with which we are calling `isExposedStatement` when iterating over `program.body`.

That doesn't work:
```
src/index.ts(52,42): error TS2345: Argument of type 'Node' is not assignable to parameter of type 'Statement'.
  Type 'AnyTypeAnnotation' is not assignable to type 'Statement'.
    Property 'id' is missing in type 'AnyTypeAnnotation' but required in type 'TSNamespaceExportDeclaration'.
```

> (I prefer to avoid the import-by-copying method since if `@babel/types` breaks again we will need to change as well so there is no real value in it).

I don't see it as import-by-copying.  This literally *is* the type we expect to see.  (If the comment type stops being exported, we're in trouble; is that what you mean?)  Anyway, I'd rather not continue to search for a good type... and I don't like the ultra-wide type `Node`, we've had bad experience with TypeScript memory usage with wide types in earlier TS versions.

arielshaqed:
Thanks!

PR #199, opened 2019-09-08
Return statusCode 204 on empty responses

vogre:
For feature parity with local-proxy


vogre:
Tested manually in this case.

We might benefit from improving the robustness of fetch-runtime https://github.com/binaris/shiftjs/blob/79387693ed44689b4a1d2dcc840c2f57f8b02b54/fetch-runtime/src/index.ts#L28 by adding a check for content-length to allow it to work against less compliant servers.

PR #200, opened 2019-09-08
Add remote E2E test

bergundy:


Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM node:10
```
arielshaqed: This shouldn't be at the top level of the repo, it belongs inside its package.  Failing that, it will need a very specific filename.

Jenkinsfile
```diff
@@ -0,0 +1,37 @@
+node {
```
arielshaqed: Shouldn't this just live in `precious`?  It anyway _says_ "precious", so it's not particularly useful to someone outside of Binaris.

app-testsuite/src/run.ts
```diff
@@ -16,6 +16,17 @@ function log(...args: any[]) {
   console.log(new Date(), ...args);
 }
 
+async function readAll(stream: NodeJS.ReadableStream) {
+  const out: string[] = [];
+  return new Promise<string>((resolve, reject) => {
+    stream.on('data', (l) => {
+      out.push(l.toString());
+    });
+    stream.on('end', () => resolve(out.join('')));
```
arielshaqed: Nit: This fails on a long UTF-8 output, if a point is split between two buffers.

cli/package.json
```diff
@@ -35,9 +35,9 @@
   "scripts": {
     "postpack": "rm -f oclif.manifest.json",
     "posttest": "tslint -c ../common/tslint.yml -p . -t stylish",
-    "prepack": "rm -rf lib && tsc -b && oclif-dev manifest && oclif-dev readme",
+    "prepack": "rm -rf lib && tsc && oclif-dev manifest && oclif-dev readme",
```
arielshaqed: Why?

cli/package.json
```diff
@@ -35,9 +35,9 @@
...
     "lint": "tslint -c ../common/tslint.yml -p .",
-    "test": "ava lib/test/**.spec.js",
+    "test": "ava lib/test/**.spec.js lib/test/**.test.js",
```
arielshaqed: I thought we standardized on `*.spec.*`?

cli/src/commands/deploy.ts
```diff
@@ -32,7 +31,7 @@ export default class Deploy extends Command {
   public async build(projectDir: string): Promise<string> {
     this.log('Building and bundling your app! This may take a few moments, please wait');
     const stagingDir = pathResolve(tmpdir(), 'shift-bundle-');
-    await mkdirp(stagingDir);
+    await mkdtemp(stagingDir, { encoding: 'utf8' });
```
arielshaqed: Yay!

cli/src/commands/list.ts
```diff
@@ -12,28 +13,51 @@ export default class List extends Command {
...
-        },
-        updatedAt: {
-          headingTransform: () => 'LAST UPDATED',
-          minWidth: 25,
-        },
-      },
+    const mappedApps = apps.map(({ name, updatedAt, environments }) => ({
+      name,
+      updatedAt: updatedAt.toISOString(),
+      URL: `https://${environments[0].domains[0].name}`,
```
arielshaqed: This is an unfortunate itnerface: if ever we change name of the app's page (a *remote* change), old CLIs will give the wrong URL.  Add another field (in another PR)?

Jenkinsfile
bergundy: That's how the flow goes for the other repos..

app-testsuite/src/run.ts
bergundy: Good point

cli/package.json
bergundy: Nope, we said, `.test.`, `.spec.` is for E2E or something AFAIR (@arikmaor)?

cli/src/commands/list.ts
bergundy: What do you mean by name of the app's page?

cli/package.json
bergundy: Just to keep it inline with the other uses of `tsc`.

Dockerfile
bergundy: Change to `testing.Dockerfile`.

cli/package.json
arikmaor: this line was created by `oclif init`

vogre: Note that the `-b` is needed for [project references](https://www.typescriptlang.org/docs/handbook/project-references.html) if we decide to use them.

bergundy: As long is it's consistent.

PR #203, opened 2019-09-08
Modify Dockefile to run cypress

bergundy:


testing.Dockerfile
```diff
@@ -1,11 +1,13 @@
 FROM node:10
 
-RUN apt update && apt install -y jq
+RUN apt update && apt install -y jq xvfb libgtk-3-dev libnotify-dev libgconf-2-4 libnss3 libxss1 libasound2
```
vogre: As long as this is cached it shouldn't be too bad.

The list comes from [cypress docs](https://docs.cypress.io/guides/guides/continuous-integration.html#Advanced-setup) so should be exactly what we need.

PR #209, opened 2019-09-09
Escape path to win32 cmd path

vogre:


vogre:
This was tested manually by making a "counter-class copy" directory on windows and deploying from there.

This is only a shell requirement:

- `spawn('/path/w i t h/spaces', ['arg with spaces'])` works correctly
- `spawn('/path/w i t h/spaces', ['arg with spaces'], { shell: true })` requires escaping

All `fs` APIs don't require escaping (although in some cases paths need to be passed as `Buffer`s instead of strings).

PR #210, opened 2019-09-10
Add dir->app selection for deploy

bergundy:


cli/src/commands/deploy.ts
```diff
@@ -117,9 +134,38 @@ export default class Deploy extends Command {
     return digest;
   }
 
+  public async selectApplicationForProject(): Promise<string | undefined> {
+    const NEW_APPLICATION = '__new_application__';
```
michaeladda: consider using something unique (though getting __new_application__ as the id is very unlikely currently)

cli/src/commands/deploy.ts
```diff
@@ -117,9 +134,38 @@ export default class Deploy extends Command {
...
+    if (value === undefined) {
+      this.error('No option selected');
+    }
+    return value === NEW_APPLICATION ? undefined : value;
+  }
+
   public async run() {
     this.parse(Deploy);
     await this.authenticate();
+    await this.selectApplicationForProject();
```
michaeladda: this should be removed

bergundy:
Thanks!

PR #213, opened 2019-09-10
Diagnose bad server responses

arielshaqed:


cli/src/utils/command.ts
```diff
@@ -95,6 +95,16 @@ export default abstract class BaseCommand extends Command {
     this.webAppLoginUrl = webAppLoginUrl;
   }
 
+  public async catch(err: Error) {
+    if (err instanceof ValidationError) {
+      this.error(
+        `Failed to communicate with server: ${err.message}.  Please run "npm update -g shift-cli".`,
```
vogre: (suggestion) This is nice for most cases, but might look silly if there's no actual update available. Perhaps we should look up the registry (with some rate-limiting of course) before we suggest this, similar to [oclif/plugin-warn-if-update-available](https://github.com/oclif/plugin-warn-if-update-available/blob/master/src/get-version.ts).

arielshaqed: Opened #214 to do this; I think this PR gives independent value and we should pull it.  Rremoving the update suggestion from the log so we can do so.

Thanks!

PR #215, opened 2019-09-11
Add support for env vars

bergundy:


app-testsuite/app/src/App.js
```diff
@@ -1,36 +1,54 @@
...
+import { get, update, getSecret } from '../backend/index';
 
-const Counter = ({ keyName }) => {
+const Stateful = ({ getter, children }) => {
   const [{ loading, data, error }, setState] = useState({ loading: true });
   useEffect(() => {
     if (loading) {
-      get(keyName)
-        .then(count => setState({ data: count || 0 }))
+      getter()
```
arielshaqed: This would be easier to read as an `async` function, I think.  Especially if ever we wanted to add something to one of the portions.

app-testsuite/app/src/App.js
```diff
@@ -1,36 +1,54 @@
...
   }
-  return (<>
-    <span className='counter'>{data}</span>
+  return children({ data, setState });
+}
+
+const CounterDisplay = ({ data, setState, keyName }) => {
+  const count = data || 0;
+  const increment = useCallback(() => {
+    update(keyName, count + 1)
```
arielshaqed: Didn't @duckranger fix this yesterday?

local-proxy/src/server.ts
```diff
@@ -25,7 +27,23 @@ if (!isAbsolute(basePath)) {
   throw new Error('SHIFT_DEV_SERVER_BASE_REQUIRE_PATH env var is not an absolute path');
 }
 
-function setupEnv({ port }: { port: number }) {
+async function loadDotEnv() {
+  const envFile = resolvePath(basePath, '.env');
```
arielshaqed: Doesn't seem like it's the right path.

bergundy: Why not?
It does work.. there's a test for it.

app-testsuite/app/src/App.js
bergundy: IDK, didn't have any conflicts

app-testsuite/app/src/App.js
bergundy: Agree, but I'd rather not delay this PR.

bergundy:
@arielshaqed it already worked for deployed apps, the jenkins integration test now verifies it.

local-proxy/src/server.ts
arielshaqed: Sorry for eliding the point and being unclear.  This is the same path as for deploying.  That _might_ be the right env, but I think I'd more often want *different* config values when running locally.  With this, I get a _separate_ reshuffle DB when I run locally but the _same_ e.g. RedisLabs DB.

bergundy:
@arielshaqed that's what the spec says

local-proxy/src/server.ts
vogre: Maybe we should re-review. An `.env` file in [heroku local](https://devcenter.heroku.com/articles/heroku-local#copy-heroku-config-vars-to-your-local-env-file) behaves in the opposite way (only affects local environment and not the remote app.

PR #217, opened 2019-09-11
Use reshuffle domain for CLI

bergundy:


bergundy:
Integration eventually passed here: https://ci.binaris.io/blue/organizations/jenkins/gobinaris%2Fprecious/detail/feature-reshuffle-domain/14/pipeline/

PR #218, opened 2019-09-12
Add download command

michaeladda:
The tar.gz url should not be calculated here, and will be derived from the template data in a future PR.


cli/src/commands/download.ts
```diff
@@ -0,0 +1,91 @@
...
+  public static description = 'download application';
+  public static examples = [
+    `$ ${Command.cliBinName} download 123456`,
+  ];
+
+  public static args = [
+    {
+      name: 'ID',
+      required: true,
+      description: 'application id',
```
arielshaqed: AFAIK other commands default to the application in the current directory.  It _would_ make sense to download my app, e.g. if I want to make sure what's running.
Can we make it optional _when_ in the directory of an application?

cli/src/commands/download.ts
```diff
@@ -0,0 +1,91 @@
...
+    },
+  ];
+
+  public static strict = true;
+
+  public async run() {
+    const { args } = this.parse(Download);
+    const { ID: applicationId } = args;
+    await this.authenticate();
+    const applications = await this.lycanClient.listApps();
```
arielshaqed: Nit: This should use some new API, `describeApp`.

cli/src/commands/download.ts
```diff
@@ -0,0 +1,91 @@
...
+    const application = applications.find((app) => app.id === applicationId);
+    if (!application) {
+      return this.error('Could not find application');
```
arielshaqed: Can we add (here and below) a "verbose" mode?  It will add details to the query (e.g. application ID, entire `application` response below, etc.  Useful for us to debug, and maybe for some power users.

cli/src/commands/download.ts
```diff
@@ -0,0 +1,91 @@
...
+    }
+    const { sourceUrl } = application;
+    if (typeof sourceUrl !== 'string') {
+      return this.error('Application source is unknown');
+    }
+    const projectBaseName = `${path.basename(sourceUrl || 'ddd')}-master`;
```
arielshaqed: `'ddd'`?

cli/src/commands/download.ts
```diff
@@ -0,0 +1,91 @@
...
+    if (project) {
+      if (applicationId !== project.applicationId) {
+        return this.error(`Directory ${projectDir} is already associated with a application ${project.applicationId}`);
+      }
+    } else {
+      projects.push(newProject);
+      this.conf.set('projects', projects);
+    }
+    this.log('Downloading application...');
+    const compressedSourceUrl = `${sourceUrl}/archive/master.tar.gz`;
```
arielshaqed: Could we "just" have a Lycan endpoint that redirects correctly (doesn't need to be in this PR)?

cli/src/commands/download.ts
```diff
@@ -0,0 +1,91 @@
...
+    // this.log(compressedSourceUrl);
+    const targetDir = '.';
+    const extract = tar.extract({ cwd: targetDir });
+    const getAppCodePromise = new Promise<void>((resolve, reject) => {
+      request.get(compressedSourceUrl)
+        .on('error', () => {
+          reject(new CLIError(`Failed fetching ${compressedSourceUrl}`));
```
arielshaqed: This loses all details of the HTTP error.  But it's interesting and helpful to the user!

cli/src/commands/download.ts
```diff
@@ -0,0 +1,91 @@
...
+        })
+        .on('response', (res) => {
+          const statusCode = res.statusCode;
+          if (statusCode !== 200) {
```
arielshaqed: Maybe test "not OK" instead?  THere are all those different 200s...  `nodeutils` has [`statusOk`](https://github.com/binaris/nodeutils/blob/master/src/http-utils.ts#L4).

cli/src/commands/download.ts
```diff
@@ -0,0 +1,91 @@
...
+    } else {
+      projects.push(newProject);
+      this.conf.set('projects', projects);
+    }
+    this.log('Downloading application...');
+    const compressedSourceUrl = `${sourceUrl}/archive/master.tar.gz`;
+    // this.log(compressedSourceUrl);
+    const targetDir = '.';
+    const extract = tar.extract({ cwd: targetDir });
+    const getAppCodePromise = new Promise<void>((resolve, reject) => {
```
arielshaqed: Nice!  But why not just `await` this `Promise`?  It would be shorter _and_ clearer than naming it, because I immediately know the promise is forced inline.

cli/src/commands/download.ts
michaeladda: indeed. this is being worked on.

cli/src/commands/download.ts
michaeladda: similar solution is being worked on (returning all download "magic" from server)

cli/package.json
```diff
@@ -65,6 +65,7 @@
     "node-fetch": "^2.6.0",
     "open": "^6.4.0",
     "prompts": "^2.2.1",
+    "request": "^2.88.0",
```
vogre: Possible to use `node-fetch` which is already installed anyway to get the result as promise, also be more [hip](https://github.com/request/request/issues/3142).

cli/src/commands/download.ts
arielshaqed: Done above, I see.  Thanks!

cli/src/commands/download.ts
```diff
@@ -0,0 +1,119 @@
...
+    const {
+      args : { ID: applicationId },
+      flags: { verbose },
+    } = this.parse(Download);
+    await this.authenticate();
+    const applications = await this.lycanClient.listApps();
+    const application = applications.find((app) => app.id === applicationId);
+    if (!application) {
+      if (verbose) {
+        this.log('All applications:', applications);
```
arielshaqed: We really need a proper log, and then always log and just change disposition of level `debug` / `verbose`.  Not in this PR.

michaeladda: totally agree!
It should also affect all commands

PR #221, opened 2019-09-14
Prevent invocation of non-exposed functions

bergundy:


local-proxy/src/server.ts
```diff
@@ -168,8 +171,15 @@ router.post('/invoke', async (ctx) => {
       }
       const mod = require(joinedDir);
       fn = mod[handler];
+      // Cast to any so typescript doesn't complain against accessing __shiftjs__ on a function
+      if (!(typeof fn === 'function' && (fn as any).__shiftjs__ && (fn as any).__shiftjs__.exposed)) {
```
vogre: Possible to extract as a common function between `server-function` and `local-proxy`.

app-testsuite/app/src/App.js
```diff
@@ -1,53 +1,77 @@
 import '@binaris/shift-code-transform/macro';
 import React, { useState, useEffect, useCallback } from 'react';
 import { get, update, getSecret } from '../backend/index';
+import { notExposed, invalidFile } from './hack';
 
-const Stateful = ({ getter, children }) => {
-  const [{ loading, data, error }, setState] = useState({ loading: true });
+console.log('NE', notExposed);
```
vogre: Nit: leftover `console`

local-proxy/src/server.ts
```diff
@@ -168,8 +171,15 @@ router.post('/invoke', async (ctx) => {
       }
       const mod = require(joinedDir);
       fn = mod[handler];
+      // Cast to any so typescript doesn't complain against accessing __shiftjs__ on a function
```
vogre: Note that this has a conflict with https://github.com/reshufflehq/reshuffle/pull/219, which renames `__shiftjs__` to `__visibility__`.

PR #228, opened 2019-09-16
Fix destroy command

michaeladda:
The destroy command was changed from `destroy ID` to `destroy  [--id ID]` recently, to destroy the project of the current directory if no ID is given. This was:
1. not documented in the example.
2. diverging from the rest of the commands (which takes args not flags. not opinionated here, but strongly prefer coherence).
3. not tested at all

This PR changes it back to `destroy [ID]` and adds tests.


cli/src/test/destroy.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+test.beforeEach(async (t) => {
+  t.assert(success(await t.context.shell.run(`cd ${t.context.projectDir}`)));
+  t.log(t.context.projectDir);
+});
+
+test('no args in project dir', async (t) => {
+  td.when(t.context.lycanFake.destroyApp(anything, 'fluffy-samaritan')).thenResolve(null);
+  const result = await t.context.shell.run(`${t.context.run} destroy`, 'utf-8');
+  t.snapshot(result);
+  const config = await readFile(t.context.configPath, 'utf-8');
```
arielshaqed: Can you snapshot here, too?

michaeladda: excellent suggestion. thanks. separates content from code.!

PR #230, opened 2019-09-18
Do not specify default in config flag

### LINKED ISSUE #225: According to CLI, everyone is Ariel
```
npx: installed 140 in 9.171s
deploy your Reshuffle project to its associated domain

USAGE
  $ reshuffle deploy

OPTIONS
  -h, --help       show CLI help
  --config=config  [default: /Users/ariels/.reshuffle/config.yml]

EXAMPLE
  $ reshuffle deploy
```
I'm not sure if its supposed to be an example ( I would add `eg:`) or just a mistake.

vogre:
Avoid bundling local build paths into manifest

Similar to cli tools like vim only show the config path in the
documentation and not in the help message

Fixes #225


common/changes/reshuffle/help-paths_2019-09-18-07-46.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "reshuffle",
+      "comment": "Remove default path from config parameter",
+      "type": "patch"
+    }
+  ],
+  "packageName": "reshuffle",
+  "email": "vladimir@shiftjs.com"
```
michaeladda: Also, you should update your `.git/config`. We've reshuffled :-p

PR #232, opened 2019-09-18
Add "rush generate-interfaces"

arielshaqed:
This is better than running `scripts/generate_interfaces` because it
is part of Rush and saves mucking around with paths.


arielshaqed:
Won't work 😢 , need generated interfaces to run `rush`.

PR #237, opened 2019-09-19
Add API-level documentation

arielshaqed:
This is auto-generated from TSDoc-like annotations on `db-client`; it gives us _something_.  You can see what on http://dev-docs.reshuffle.com/ (which is also how I tested it).

It turns out that there are no good simple documentation tools for tsdoc.  We use TypeDoc, which is the best I could find with minimal setup.  It has limitations; we shall need either to work around or fix them.

Alternatives considered:
1. Microsoft do The Real TsDoc:tm:.  This will be great, but is not yet: they have a parser, but haven't even documented the format, and there is no "publish docs from this directory" tool.  Also the format is more limited than that for typedoc.  There are additional tools based on API Extractor, but I was unable to get them to work.
2. Angular have dgeni; this is what's used by RxJS for their pretty smooth website.  But it's really a library and you have to put together the documenter you want; more than 2 days' work (I tried).
3. JSDoc is hard to use with TypeScript.

### DIFF
--- a/common/changes/@reshuffle/db/docs_2019-09-19-14-47.json
+++ b/common/changes/@reshuffle/db/docs_2019-09-19-14-47.json
@@ -3,7 +3,7 @@
     {
       "packageName": "@reshuffle/db",
       "comment": "Generate documentation using typedoc",
-      "type": "patch"
+      "type": "none"
     }
   ],
   "packageName": "@reshuffle/db",
--- a/common/changes/@reshuffle/db/docs_2019-09-19-14-47.json
+++ b/common/changes/@reshuffle/db/docs_2019-09-19-14-47.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@reshuffle/db",
+      "comment": "Generate documentation using typedoc",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@reshuffle/db",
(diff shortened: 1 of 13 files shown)


arielshaqed:
Oh yes: this is deployed in practice -- but not linked to.

ashevat:
I need to comment on the content created by this auto-generation - not the code itself. What is the best way to do that?

arielshaqed:
I need to comment on the content created by this auto-generation - not the
> code itself. What is the best way to do that?
>

It's part of the reshuffle sources, on the function documentation
(subdirectory db-client/src/, but I'm not sure you need to care). You're
right, it's important but out of scope for this PR.

Please open an issue in the repo.

common/changes/@reshuffle/db/docs_2019-09-19-14-47.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@reshuffle/db",
+      "comment": "Generate documentation using typedoc",
+      "type": "patch"
```
vogre: Can be a `none` type since the package is unchanged, the documentation does not become a part of the package.

arielshaqed:
Thanks!

I opened items for doc fixes and terraform, and pulling.

PR #243, opened 2019-09-22
Store all temporary files in .reshuffle dir

bergundy:


local-proxy/src/server.ts
```diff
@@ -91,13 +93,13 @@ const app = new Koa();
 app.use(bodyParser({ enableTypes: ['json'], strict: false }));
 const router = new KoaRouter();
 
-const tmpDir = mkdtempSync(resolvePath(basePath, '..', '.reshuffle_local_proxy_'));
+const genDir = mkdtempSync(resolvePath(tmpDir, 'local_proxy_'));
```
vogre: This has an issue, that now it's difficult to reliably access a file from `src` in code that is in `backend`, since it will have a different relative path.

This might not be a problem yet, since none of our apps currently do this, but to allow it I think we can add a new variable `RESHUFFLE_PROJECT_ROOT` or something similar.

local-proxy/src/server.ts
```diff
@@ -14,11 +14,13 @@ import { AddressInfo } from 'net';
 import os from 'os';
 import { initRegistry } from './stdio';
 import nanoid from 'nanoid';
-import mkdirp from 'mkdirp';
-import { copy } from 'fs-extra';
+import { copy, mkdirpSync } from 'fs-extra';
```
vogre: `mkdirp` can now be removed from `package.json` dependencies.

Similarly possible to change calls to `rimraf`.

local-proxy/src/server.ts
```diff
@@ -14,11 +14,13 @@ import { AddressInfo } from 'net';
...
 import env from 'env-var';
 import dotenv from 'dotenv';
 
+const tmpDir = env.get('RESHUFFLE_TMP_DIR').required().asString();
```
vogre: Can still be checked for `isAbsolute()`, just in case.

bergundy: Does it matter?

local-proxy/src/server.ts
bergundy: Good catch!
I'm not sure we want to let users import from `src`, we don't even upload it today as part of the bundle.

https://github.com/reshufflehq/reshuffle/blob/2f5854729d75bd2c71d14d2f5652494d17a0713c/cli/src/commands/deploy.ts#L73

local-proxy/src/server.ts
vogre: If it's e.g. `./.reshuffle` and the code does `process.chdir()` somewhere it suddenly breaks.

bergundy: It's not possible ATM since we use `tmpDir` with `resolvePath` before requiring user code.
We might want to `resolvePath` immediately for `tmpDir` to prevent future errors.

vogre: It's not possible ATM, but so are all other environment variable checks but it comes to protect from 99% of mistakes where an `undefined` or empty string are passed.

PR #247, opened 2019-09-23
Don't use @binaris/utils-subprocess (it's also now @reshuffle'd)

arielshaqed:


arielshaqed:
Thanks!

Will pull once integration passes (that *is* the point, here).

arielshaqed:
Passed Jenkins CI in https://ci.binaris.io/blue/organizations/jenkins/gobinaris%2Fprecious/detail/master/8838, so Admin Force Pull!

PR #248, opened 2019-09-23
Get @binaris/ scoped packages from npmjs

vogre:
We no longer publish packages with @binaris from reshuffle repo
Used for @binaris/spice-node-client

Continuation of #247


vogre:
Alredy part of #247.

PR #251, opened 2019-09-23
Drop detailed Binaris logs unless `--all` is specified

arielshaqed:
`--all` keeps the old behaviour: all logs emitted.

Without it, function deployment and invocation logs are dropped.  Note
that binaris/spice#677 adds _application_ logs, which do show up.


cli/src/test/snapshots/logs.spec.ts.md
```diff
@@ -4,59 +4,6 @@ The actual snapshot is saved in `logs.spec.ts.snap`.
 
 Generated by [AVA](https://ava.li).
 
-## logs
```
michaeladda: i don't understand. how can the tests pass if the md missing old tests and the new ones are missing?

cli/src/commands/logs.ts
```diff
@@ -6,13 +6,21 @@ import {
 } from '../utils/helpers';
 import ms = require('ms');
 
+const detailedLogsRegexps = [
+  /Function invocation took [\d.]+ us/,
```
michaeladda: Shouldn't this be `/^Function ...` (or perhaps the pattern is so rare than it's not interesting?)

arielshaqed: You're right, I didn't anchor it.

arielshaqed:
> I think there's no positive match for the second regex in the tests
> (it's always partial so no match)

True, thanks!  Guess what?  They were both slightly wrong, as were some supposedly triggering messages.

Will pull once tests pass.

PR #258, opened 2019-09-23
Use download target dir

michaeladda:
both unit tested and manually
directory override matching git clone behaviour


cli/src/commands/download.ts
```diff
@@ -106,14 +106,15 @@ export default class Download extends Command {
           reject(new CLIError(`Failed fetching ${downloadUrl}`));
         });
     });
+    await rename(downloadDir, targetDir);
```
vogre: If this throws we need to clean up the download?

cli/src/commands/download.ts
```diff
@@ -78,8 +79,7 @@ export default class Download extends Command {
       this.conf.set('projects', projects);
     }
     this.log('Downloading application...');
-    const targetDir = '.';
-    const extract = tar.extract({ cwd: targetDir });
+    const extract = tar.extract({ cwd: '.' });
```
vogre: Might be easier check before with `fs-extra` `emptyDir` to avoid the whole download just to fail at rename. (still a race of course).

cli/src/commands/download.ts
```diff
@@ -78,42 +81,73 @@ export default class Download extends Command {
...
+            let first = true;
+            res.body.pipe(extract)
+              .on('error', (err) => {
+                if (first) {
+                  verboseLog('extract', err);
+                  first = false;
+                }
+                reject(new CLIError('Failed extracting application'));
+              })
+              .on('finish', () => {
```
vogre: It seems like the correct event should be `close` in latest `tar` or `minipass` (see https://github.com/binaris/valet/pull/412).

cli/src/test/download.spec.ts
```diff
@@ -16,6 +16,15 @@ const defaultApp = {
   environments: [],
 };
 
+const stabalize = (result: Output<string> & (Exit | Signal)) => ({
```
vogre: Nit: stabilize?

cli/src/commands/download.ts
```diff
@@ -78,42 +81,73 @@ export default class Download extends Command {
...
-              reject(new CLIError('Failed extracting application'));
-            })
-            .on('finish', () => {
-              resolve();
-            });
-        })
-        .catch((err) => {
-          verboseLog('download', err);
-          reject(new CLIError(`Failed fetching ${downloadUrl}`));
+    try {
```
vogre: Possible to restructure a bit with early return by
```
let res: Response;
try {
  res = await fetch(downloadUrl);
} catch (e) {
  verboseLog('download', err);
  throw new CLIError(`Failed fetching ${downloadUrl}`));
}
if (statusNotOk(statusCode)) {
  throw new CLIError(`Bad status code ${statusCode} when fetching ${downloadUrl}`);
}
await new Promise((resolve, reject) => {
    res.body.pipe(extract)...

});
```

PR #259, opened 2019-09-23
Made some low-priority readability changes

duckranger:


duckranger:
Did not have rush configured.

arielshaqed:
> Text clarification looks great.
> However, please run `rush change` (after installing it locally) and add the generated changelog file to this PR. Without it, automatic versioning will not work, and the CI will fail

Note this is a documentation fix, so patch (no new functionality).  Please ping me after you're done pulling and I'll republish docs -- it's manual for now.

common/changes/@reshuffle/db/grammar-suggestions_2019-09-25-06-34.json
```diff
@@ -0,0 +1,11 @@
+{
```
arielshaqed: You could avoid this change entry by hitting ENTER instead of entering text when `rush change` asks.  This is change does not need a new version number AFAICS.

vogre: Note that a `none` change would only be published with later patches, so anyone using the package with vscode will not see the documentation before the next publish, so it's a tradeoff between how often we want to update.

michaeladda:
merging, despite not marked as "merge if approved". @duckranger, for future reference, when we want the reviewer to merge on her own we add such label to the PR :-)

PR #260, opened 2019-09-24
Prettify LevelDB errors correctly

### LINKED ISSUE #255: leveldb-server: Cannot assign to read only property 'message' of object 'ReadError: Database is not open'
https://github.com/reshufflehq/reshuffle/blob/0cc6d09f960107ea1199eb4cb57a836e1cb52c17/leveldb-server/src/db.ts#L129

Reproducible when running `npm start` in the same directory from two different terminals - error should highlight that database is already open.

arielshaqed:
Don't touch immutable `Error`s, copy them instead.  Resulting error
looks like this:
```
Failed to invoke function { InternalServerError: [6hMvBxG47sy90hXscv-DJ] Database is not open
    at DBClient.get (/Users/ariels/Google Drive/Work/Dev/counter-class-master/node_modules/@reshuffle/interfaces-node-client/client.js:108:19)
    at process._tickCallback (internal/process/next_tick.js:68:7) name: 'InternalServerError' }

  InternalServerError: [6hMvBxG47sy90hXscv-DJ] Database is not open
      at DBClient.get (/Users/ariels/Google Drive/Work/Dev/counter-class-master/node_modules/@reshuffle/interfaces-node-client/client.js:108:19)
      at process._tickCallback (internal/process/next_tick.js:68:7)
```

Fixes #255.

Tested (see above) by manually linking and running 2 conncurrent local
servers.

However I *cannot* write a (simple) unit test: as detailed in #255,
when opening a LevelDB server fails there is a *previous* asynchronous
error that cannot (easily?) be caught an processed.  So any test fails
claiming e.g.
```
  /Users/ariels/Google Drive/Work/Dev/reshuffle/common/temp/node_modules/.registry.npmjs.org/levelup/4.2.0/node_modules/levelup/lib/levelup.js:96

  OpenError: IO error: lock /var/folders/k7/8lqpw3ws3ybd5_rm_1jvzlj00000gn/T/test-state-vSftNV/root.db/LOCK: already held by process

  /Users/ariels/Google Drive/Work/Dev/reshuffle/common/temp/node_modules/.registry.npmjs.org/levelup/4.2.0/node_modules/levelup/lib/levelup.js:96:23
  /Users/ariels/Google Drive/Work/Dev/reshuffle/common/temp/node_modules/.registry.npmjs.org/abstract-leveldown/6.1.1/node_modules/abstract-leveldown/abstract-leveldown.js:30:14
  /Users/ariels/Google Drive/Work/Dev/reshuffle/common/temp/node_modules/.registry.npmjs.org/deferred-leveldown/5.2.1/node_modules/deferred-leveldown/deferred-leveldown.js:22:21
  /Users/ariels/Google Drive/Work/Dev/reshuffle/common/temp/node_modules/.registry.npmjs.org/abstract-leveldown/6.1.1/node_modules/abstract-leveldown/abstract-leveldown.js:30:14
```
This comes straight outta the event loop, so not possible to catch.

(I *could* start the server in a separate process.  I claim this is
too much cruft for the size of this fix -- but a reviewer can change
this!)

### DIFF
--- a/leveldb-server/src/db.ts
+++ b/leveldb-server/src/db.ts
@@ -80,7 +80,7 @@ export class Handler implements DBHandler {
   protected readonly writeLock = new Mutex();
   protected readonly db: LevelUp;
 
-  constructor(protected readonly dbPath: string, errorCallback?: (err: Error) => any) {
+  constructor(protected readonly dbPath: string, errorCallback?: (err: Error | undefined) => any) {
     this.db = new LevelUpCtor(new LevelDown(dbPath), errorCallback);
   }
 
--- a/leveldb-server/src/db.ts
+++ b/leveldb-server/src/db.ts
@@ -30,6 +30,13 @@ export type KeyedPatches = Array<[string, Patch[]]>;
 const NUM_PATCHES_TO_KEEP = 20;
 const DEFAULT_READ_BLOCK_TIME_MS = 50000;
 
+export class WrappedError extends Error {
+  constructor(msg: string, public readonly err: Error, public readonly debugId: string) {
+    super(msg);
+    this.stack += `\ncaused by:\n${err.stack}`;
+  }
+}
+
 export function incrVersion({ major, minor }: Version, amount: number = 1): Version {
   return { major, minor: minor + amount };
 }
@@ -72,8 +79,9 @@ export class Handler implements DBHandler {
   protected emitter = new EventEmitter();
   protected readonly writeLock = new Mutex();
   protected readonly db: LevelUp;
-  constructor(protected readonly dbPath: string) {
-    this.db = new LevelUpCtor(new LevelDown(dbPath));
+
+  constructor(protected readonly dbPath: string, errorCallback?: (err: Error) => any) {
+    this.db = new LevelUpCtor(new LevelDown(dbPath), errorCallback);
   }
 
   public async extractContext(_ctx: KoaContext): Promise<ServerOnlyContext> {
@@ -125,7 +133,7 @@ export class Handler implements DBHandler {
       return JSON.parse(val.toString());
     } catch (err) {
       if (err.notFound) return undefined;
-      throw { ...err, debugId, message: `[${debugId}] ${err.message}` };
+      throw new WrappedError(`[${debugId}] ${err.message}`, err, debugId);
     }
   }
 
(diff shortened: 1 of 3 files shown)


leveldb-server/src/db.ts
```diff
@@ -125,9 +125,7 @@ export class Handler implements DBHandler {
       return JSON.parse(val.toString());
     } catch (err) {
       if (err.notFound) return undefined;
-      err.debugId = debugId;
-      err.message = `[${debugId}] ${err.message}`;
-      throw err;
+      throw { ...err, debugId, message: `[${debugId}] ${err.message}` };
```
vogre: Although this is legal, I suggest we avoid `throw`ing non-`instanceof Error`s and have a tiny `class DBError extends Error` for all `toString()`, `stack` goodiness.

vogre:
Note that test failures should go away after rebasing (see #258).

leveldb-server/src/db.ts
arielshaqed: Here's the thing.  If I add an `Error`, its constructor generates a new stacktrace -- which is all the boring bits.  And this isn't Java, we don't have proper exception chaining.
So I'm doing it using _fake_ exception chaining.  It's not like JavaScript `Error`s have any useful fields other than `message`, `type` and `stacktrace`.

leveldb-server/src/db.ts
```diff
@@ -72,8 +79,9 @@ export class Handler implements DBHandler {
   protected emitter = new EventEmitter();
   protected readonly writeLock = new Mutex();
   protected readonly db: LevelUp;
-  constructor(protected readonly dbPath: string) {
-    this.db = new LevelUpCtor(new LevelDown(dbPath));
+
+  constructor(protected readonly dbPath: string, errorCallback?: (err: Error) => any) {
```
vogre: Note the lint failure highlights that types of err here is missing undefined (the levelup type is `(err: Error | undefined) => void`).

arielshaqed: Thanks!

Strange the compiler didn't catch it, I'm passing an `errorCallback` that cannot handle `undefined` inputs to a function that says it might pass `undefined`.  Stranger still, while playing around with it I'm pretty sure I saw `null` being passed.  We shall see.

vogre:
`tsc` allows it since one overload of the constructor is `db: LevelDown, options: any, cb?: ErrorCallback`, to force it to recognize `cb` possible to pass `{}` as options and remove ambiguity.

PR #261, opened 2019-09-24
Insert initial data locally if db not initialized 

vogre:
This is probably missing some edge cases, but has the behavior in place

### DIFF
--- a/local-proxy/src/server.ts
+++ b/local-proxy/src/server.ts
@@ -205,7 +205,10 @@ if (!existsSync(dbPath) && existsSync(initDataPath)) {
       throw new Error('initData.items');
     }
   } catch (err) {
-    // just ignore any errors
+    // Logging during startup will usually get cleared by create-react-app
+    // screen clear, but the log will still be saved in the log directory
+    // tslint:disable-next-line:no-console
+    console.error('Error processing template_init_data.json', err);
   }
 }
 const db = new DBHandler(dbPath, initData);
--- a/local-proxy/src/server.ts
+++ b/local-proxy/src/server.ts
@@ -1,5 +1,5 @@
 import { promisify } from 'util';
-import { resolve as resolvePath, extname, isAbsolute } from 'path';
+import { resolve as resolvePath, extname } from 'path';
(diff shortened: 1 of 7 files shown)


local-proxy/src/server.ts
```diff
@@ -192,7 +195,20 @@ router.post('/invoke', async (ctx) => {
   }
 });
 
-const db = new DBHandler(resolvePath(tmpDir, 'db'));
+const dbPath = resolvePath(tmpDir, 'db');
+const initDataPath = resolvePath(rootPath, 'template_init_data.json');
```
michaeladda: not for this PR: the name of the file should probably come from lycan interface so there's only one source of truth

local-proxy/src/server.ts
```diff
@@ -192,7 +195,20 @@ router.post('/invoke', async (ctx) => {
...
+let initData: any;
+if (!existsSync(dbPath) && existsSync(initDataPath)) {
+  try {
+    initData = JSON.parse(readFileSync(initDataPath, 'utf8'));
+    if (!Array.isArray(initData.items)) {
+      throw new Error('initData.items');
+    }
+  } catch (err) {
+    // just ignore any errors
```
michaeladda: not for this PR: please print

local-proxy/src/server.ts
```diff
@@ -192,7 +195,20 @@ router.post('/invoke', async (ctx) => {
   }
 });
 
-const db = new DBHandler(resolvePath(tmpDir, 'db'));
+const dbPath = resolvePath(tmpDir, 'db');
+const initDataPath = resolvePath(rootPath, 'template_init_data.json');
+let initData: any;
+if (!existsSync(dbPath) && existsSync(initDataPath)) {
+  try {
+    initData = JSON.parse(readFileSync(initDataPath, 'utf8'));
```
michaeladda: not for this PR: consider extracting the processing logic (schema validation iirc) to shared place?

local-proxy/src/server.ts
vogre: Do you think the source of truth should be top-down server->client, because thinking about it now it will be easier to reverse and export it from the client?

PR #263, opened 2019-09-25
Update spice client to exclude name from deployInitial

bergundy:
Goes with the discussed changes to spice


cli/src/commands/deploy.ts
```diff
@@ -12,7 +12,6 @@ import { Application } from '@binaris/spice-node-client/interfaces';
 import Command from '../utils/command';
 import { getDependencies } from '../utils/getdeps';
 import {
-  getProjectPackageJson,
```
michaeladda: i think `getProjectPackageJson` is now unused in helps so should go away

bergundy:
Integration failure unrelated.

PR #265, opened 2019-09-25
Skip building ephemeral publish- branch

vogre:
On Jenkins in addition to CircleCI

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,8 +16,7 @@ node {
     ]);
 
     if ("${BRANCH_NAME}" =~ "^publish-.*") {
-        currentBuild.result = 'SUCCESS'
-        return
+        return;
     }
 
     genericWrapper([slackChannel:'#jenkins', timeoutMinutes: 60]) {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -15,6 +15,11 @@ node {
         ]),
     ]);
 
+    if ("${BRANCH_NAME}" =~ "^publish-.*") {
+        currentBuild.result = 'SUCCESS'
```
bergundy: Is this required?

Jenkinsfile
```diff
@@ -15,6 +15,11 @@ node {
...
+        return
```
bergundy: Can haz semicolons?

Jenkinsfile
vogre: Seems to work without it.

Jenkinsfile
vogre: https://dzone.com/articles/groovy-sometimes-you-still

Groovy has fewer edge cases, but I will never remember them either.

PR #268, opened 2019-09-26
Add support for logs by app name

michaeladda:


cli/src/commands/logs.ts
```diff
@@ -62,27 +65,50 @@ $ ${Command.cliBinName} logs --since 2m --follow`,
     }),
   };
 
-  public static args = [];
+  public static args = [
+    {
+      name: 'name',
+      required: false,
+      description: 'Application name (defaults to working directory\'s deployed application name)',
```
arielshaqed: (Not this PR)
It would be cool if I could run some command to determine this.  WDYT?

cli/src/commands/logs.ts
```diff
@@ -62,27 +65,50 @@ $ ${Command.cliBinName} logs --since 2m --follow`,
...
+      args: { name: appName },
+    } = this.parse(Logs);
     await this.authenticate();
 
-    const projectDir = await getProjectRootDir();
-    const projects = this.conf.get('projects') as Project[] | undefined || [];
-    const project = projects.find(({ directory }) => directory === projectDir);
-    if (project === undefined) {
-      return this.error(`No project deployments found, please run ${Command.cliBinName} deploy`);
+    let applicationId: string | undefined;
```
arielshaqed: These lines could be in a separate function.

cli/src/test/logs.spec.ts
```diff
@@ -100,3 +101,28 @@ test('logs help', async (t) => {
   const result = await t.context.shell.run(`${t.context.run} help logs`, 'utf-8');
   t.snapshot(result);
 });
+
+test('logs by name missing app', async (t) => {
+  td.when(t.context.lycanFake.getAppByName(anything, 'not-here'))
+    .thenReject(new NotFoundError('not found'));
+
+  const result = await t.context.shell.run(`cd .. && ${t.context.run} logs not-here`, 'utf-8');
```
arielshaqed: Nit: `cd ..` might stay in the same directory, e.g. if we run in the root directory (of a container, presumably).  Can we `mkdir pirate && cd pirate && ...` instead?

(Also below)

cli/src/commands/logs.ts
michaeladda: makes sense. maybe even `reshuffle describe` which would describe the current app or the app by name

PR #270, opened 2019-10-02
[CLI/deploy] Verify package-lock contains all packages to install

### LINKED ISSUE #264: Deploy fails in case of mismatched package-lock.json
https://github.com/reshufflehq/reshuffle/blob/1d527534ac95bc792297fa67576e8049238e689d/cli/src/utils/getdeps.ts#L24

sez

```
TypeError: Cannot read property 'requires' of undefined
    at Object.getDependencies (~/dev/reshuffle/cli/src/utils/getdeps.ts:24:54)
```

Steps to reproduce

Install a random npm package
Restore package-lock.json to previous state

arielshaqed:
Otherwise there is no way to deploy that package.

Fixes #264.

Unfortunately we have no deploy test yet, so tested manually by repeating steps from the issue:
```
flutterby:app ariels$ ~/Dev/reshuffle/cli/bin/run deploy
Building and bundling your app! This may take a few moments, please wait

> counter-class@0.1.2 build /Users/ariels/Google Drive/Work/Dev/counter-class-master
> react-scripts build

Creating an optimized production build...
Compiled successfully.

File sizes after gzip:

  41.56 KB  build/static/js/2.258bc445.chunk.js
  1 KB      build/static/js/main.1edd228b.chunk.js
  762 B     build/static/js/runtime~main.a8a9905a.js
  438 B     build/static/css/main.8f98742a.chunk.css

The project was built assuming it is hosted at the server root.
You can control this with the homepage field in your package.json.
For example, add this to build it for GitHub Pages:

  "homepage" : "http://myname.github.io/myapp",

The build folder is ready to be deployed.
You may serve it with a static server:

  npm install -g serve
  serve -s build

Find out more about deployment here:

  https://bit.ly/CRA-deploy

Preparing backend...
 ›   Error: Mismatched package files: package.json refers to kurkar, which is 
 ›   not in package-lock.json.  Re-run "npm install".
```


cli/src/utils/getdeps.ts
```diff
@@ -1,5 +1,12 @@
 import { resolve as pathResolve } from 'path';
 import { readFile } from 'mz/fs';
+import { CLIError } from '@oclif/errors';
+
+class MismatchedPackageAndPackageLockError extends CLIError {
+  constructor(public missingPackage: string) {
+    super(`Mismatched package files: package.json refers to ${missingPackage}, which is not in package-lock.json.  Re-run "npm install".`);
```
vogre: Nit: single space after period?

cli/src/utils/getdeps.ts
```diff
@@ -8,8 +15,18 @@ export async function getDependencies(projectDir: string) {
   const { dependencies: lockDeps } = JSON.parse(await readFile(packageLockPath, 'utf8'));
   const { dependencies: pkgDeps } = JSON.parse(await readFile(packageJsonPath, 'utf8'));
 
-  const dependencies = new Set<string>();
   const toProcess = Object.keys(pkgDeps);
+
+  // If package.json contains packages that are not in
+  // package-lock.json then they are out of sync.  Complain.
+  const lockHasProperty = Object.prototype.hasOwnProperty.bind(lockDeps);
```
vogre: I don't really like `Object.prototype.hasOwnProperty` (since it's so long to read and write). IMO something like `lodash.has` is at least much shorter to read (even if the implementation is the same).

I think we should avoid trying to prototype-proof _all_ the code, there will always be a piece of code that behaves in strange ways when encountering unusual objects. This code _would_ correctly protect against the case if we installed [prototypes](https://www.npmjs.com/package/prototypes) and then running on a `package.json` that has a dependency on [concat](https://www.npmjs.com/package/concat), but since we have some control over our dependencies we can try not to ones that modify the object prototype.

PR #273, opened 2019-10-06
Download into local directory, not tmp

arielshaqed:
Fixes these issues (due to @vogre, no actual user-facing issues filed):
1. When `/tmp/` (or whatever temporary directory is used) is on
       another directory, there is _no_ atomic rename from the temporary
       directory into the current directory.
2. Write permissions on the temporary directory were needed.

This may fix (or at least aid) using `[cli] deploy` on Windows.


arielshaqed:
Thanks!
Pulling in 2 minutes, once Jenkins passes

PR #276, opened 2019-10-07
Disable copy of node_modules

vogre:
Conflicts with the node_modules transpiled by babel


michaeladda:
@vogre What's the status of this PR? Do we want this or is it only for the time backend will have separate package.json?

vogre:
Probably requires a more complete design, closing for now since it's not functional yet.

PR #279, opened 2019-10-10
Hide raw commands

michaeladda:


cli/src/test/download.spec.ts
```diff
@@ -20,6 +20,7 @@ const stabilize = (result: Output<string> & (Exit | Signal)) => ({
   ...result,
   out: result.out
     .replace(/up to date in [0-9][.0-9]*s/g, 'up to date')
+    .replace(/found [0-9]* vulnerabilities/g, '')
```
vogre: Can there be more than 0 vulnerabilities for an empty package.json that we use in test?

PR #285, opened 2019-10-22
Add --exclude-packages flag to exclude some packages from auto-update 

vogre:
Exclude concord generated/generating packages from auto-update

To allow PRs like #284 to work again.


vogre:
Note: this must be a CLI flag and not a property in `rush.json`, since `rush.json` is JSONSchema validated and can't have additional properties.

PR #286, opened 2019-10-23
Allow PUBLIC_URL to be undefined

vogre:
PUBLIC_URL only defined in create-react-app, assume / if undefined


vogre:
This is for future experiments.

PR #291, opened 2019-10-29
Allow user to override local-proxy HTTP handler

bergundy:
This is obviously a hack but I couldn't think of a better way to solve this according to the spec.


bergundy:
@vogre please note that this is not working yet, need to pass the integration test.

bergundy:
Will reopen from feature branch

PR #295, opened 2019-10-31
Fallback to index.html in serveBinaris

bergundy:


server-function/src/serveBinaris.ts
```diff
@@ -58,6 +58,7 @@ app.post('/invoke', express.json(), async (req, res) => {
 });
 
 app.use(express.static(buildDir));
+app.use((_req, res) => res.sendfile(pathResolve(buildDir, 'index.html')));
```
vogre: Can you make sure content-type / charset match the ones served by static serving?

app-testsuite/cypress/integration/full_spec.js
```diff
@@ -28,8 +28,7 @@ describe('hack', () => {
 
 describe('user HTTP handler', () => {
   it('gets from express handler', () => {
-    cy.visit('/');
+    cy.visit('/express');
```
vogre: Cool, now it'll be tested.

server-function/src/serveBinaris.ts
vogre: Express res.sendfile seems deprecated and recommends res.sendFile instead

Internally it's implemented yhe same as static so content-type should be correct

bergundy: `Content-Type: text/html; charset=UTF-8`, it's the same.
Changing to `sendFile` and on `GET` requests.

Issue #294, opened 2019-10-29
Directory comparison on windows should be case-insensitive


vogre:
https://github.com/reshufflehq/reshuffle/blob/8db759cb9d424fc71e5ea24c087aa3c9e075bb8c/cli/src/commands/deploy.ts#L194

Drive letter can be `C:` or `c:` depending on `cwd`

arielshaqed:
Also, for added lulz, on some (but not all) Macs is case insensitive APFS. So there is no simple path comparison. I think the best you can do is `stat` each path and look

bergundy:
All directory comparisons should go through this function:
https://github.com/reshufflehq/reshuffle/blob/8aab567eb0ab4a5e4ecd6433d05644dccac7101e/cli/src/utils/helpers.ts#L13

PR #297, opened 2019-10-31
Add reshuffle rename command

bergundy:


cli/src/commands/rename.ts
```diff
@@ -0,0 +1,72 @@
...
+          await this.lycanClient.renameApp(appIdOrName, targetName);
+        } catch (err) {
+          if (err.name === 'NotFoundError') {
+            throw new CLIError('The current application is not deployed');
+          }
+          throw err;
+        }
+      }
+    } catch (error) {
+      // TODO: add verbose logging of the entire error
```
vogre: It should be possible to use `this.debug` here.

cli/src/commands/rename.ts
```diff
@@ -0,0 +1,72 @@
...
+import flags from '../utils/cli-flags';
+import {
+  getProjectRootDir,
+  findProjectByDirectory,
+} from '../utils/helpers';
+
+export default class Rename extends Command {
+  public static description = 'rename an application';
+  public static examples = [
+    `$ ${Command.cliBinName} rename great-rhinoceros-42 surpassing-lemur-17`,
```
arielshaqed: This help doesn't give the direction; consider adding some help text to this example.  Assembly language ha{d,s} this problem with AT&T (?) syntax, best be sure.  Actually I'm not even sure of the relationship this arg has with the `-s` flag argument, and O'Clif is too confusing for me to find these 2 positional arguments.

cli/src/utils/helpers.ts
```diff
@@ -10,6 +10,14 @@ export interface Project {
   defaultEnv: string;
 }
 
+export function findProjectByDirectory(projects: Project[] | undefined, projectDir: string) {
+  if (projects === undefined) {
+    return undefined;
+  }
+  // TODO: case insensitive search dependening on OS
```
arielshaqed: Nit: Please link to #294.

PR #299, opened 2019-11-03
Add express variables (url, originalUrl, baseUrl) test

vogre:


vogre:
I agree that the test does nothing more than test the functionality of express.js, but having it here validates that our local and remote implementations behave the same. Once we remove the workarounds we can remove this test, since it won't provide any new information.

PR #300, opened 2019-11-05
Fix usage of express.json() in user HTTP handler

bergundy:


server-function/package.json
```diff
@@ -15,11 +15,11 @@
   "author": "",
   "license": "MIT",
   "dependencies": {
+    "@types/express": "^4.17.1",
```
vogre: Is there a reason to push `@types/express` as dependency? This will install `@types` package for non-typescript users... Can it be /// referenced?

bergundy: Didn't think of that option.
Don't think it matters too much.

PR #307, opened 2019-11-07
Enable source map support

vogre:
This change puts the source-map-support in the uploaded serveBinaris
file, so that the support is enabled in both local and remote runtimes

The price of this change is upload of source-map(-support) which adds
944K (230K compressed) to the upload


vogre:
Thanks!

Added test and discovered an additional issue: https://www.notion.so/reshuffle/Source-maps-not-preserved-in-double-transpilation-e-g-TypeScript-JS-9785156353de43c3a9eb25ce0359fbce

PR #313, opened 2019-11-12
Use keepalive in DB client connections

bergundy:


bergundy:
@arielshaqed @vogre and I noticed yesterday while running `strace` on a running instance that without the `keepAlive` option the client creates a new HTTPS connection for every request requiring expensive SSL handshakes.
Adding `keepAlive` seemed to mitigate this issue.

arielshaqed:
Sgtm. You still probably need tls sessions, but that can wait for later. (That sessions require manual control is a bit unfortunate).
And let's also consider making this easy on concord, it should probably be the default.

PR #317, opened 2019-11-14
Add reshuffle-local-server helper script

vogre:
For running without create-react-app
Used in a package.json as { "start": "reshuffle-local-server" }

### DIFF
--- a/app-testsuite/src/run.ts
+++ b/app-testsuite/src/run.ts
@@ -1,12 +1,15 @@
 // tslint:disable:no-console
 import path from 'path';
-import { mkdtemp, readFile } from 'mz/fs';
+import { mkdtemp, readFile, writeFile } from 'mz/fs';
 import { tmpdir } from 'os';
 import { spawn as spawnChild, ChildProcess } from 'child_process';
 import { copy, remove, pathExists } from 'fs-extra';
 import { Tail } from 'tail';
 import sleep from 'sleep-promise';
 import { spawn, waitOnChild } from '@reshuffle/utils-subprocess';
+import got from 'got';
+import _assert from 'assert';
+const assert = _assert.strict;
 
 const shell = process.platform === 'win32';
 const ROOT_DIR = path.resolve(__dirname, '..', '..');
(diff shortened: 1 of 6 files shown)


local-proxy/src/localServer.ts
```diff
@@ -0,0 +1,33 @@
...
+import { setupProxy } from './index';
+import program from 'commander';
+const NAME = 'reshuffle-local-server';
+program.option('-p, --port', 'listening port (3000 by default)');
+program.description(`Run the ${NAME} outside of create-react-app server`);
+
+function main() {
+  program.parse(process.argv);
+  // create the express app to route local requests
+  const app = express();
```
bergundy: There should be a way for the user to configure his own routes and middleware for this to be useful.

local-proxy/src/localServer.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  // reshuffle currently assumes running from src and uses
+  // a relative path resolution to reach backend
+  setupProxy(process.cwd() + '/src')(app);
```
bergundy: Does this work on windows? (just verifying)

local-proxy/src/localServer.ts
bergundy: Scratch this comment

bergundy:
Can we add tests for this?

app-testsuite/src/run.ts
```diff
@@ -1,12 +1,15 @@
...
+import { mkdtemp, readFile, writeFile } from 'mz/fs';
 import { tmpdir } from 'os';
 import { spawn as spawnChild, ChildProcess } from 'child_process';
 import { copy, remove, pathExists } from 'fs-extra';
 import { Tail } from 'tail';
 import sleep from 'sleep-promise';
 import { spawn, waitOnChild } from '@reshuffle/utils-subprocess';
+import got from 'got';
+import _assert from 'assert';
+const assert = _assert.strict;
```
bergundy: Or just `import { strict as assert } from 'assert';

local-proxy/src/localServer.ts
bergundy: @vogre PING

vogre: Yep! It does produce a strange path at disk root (`C://\src`, if you decide that your project root is on `C:\\`, but even that works).

PR #319, opened 2019-11-17
Fix package lock parsing for nested deps

bergundy:
Tested on next demo app and marketplace-cms koa

### DIFF
--- a/cli/src/utils/getdeps.ts
+++ b/cli/src/utils/getdeps.ts
@@ -47,7 +47,7 @@ export async function getDependencies(projectDir: string) {
 
     if (lockHasProperty(p)) {
       dependencies.add(p);
-      toProcess.push(...recurseRequires(lockDeps[p]));
+      toProcess.push(...deepRequires(lockDeps[p]));
     } else {
       // tslint:disable-next-line:no-console
       console.error(`WARN: Cannot find dependency ${p} in package-lock.json, skipping upload`);
@@ -56,13 +56,13 @@ export async function getDependencies(projectDir: string) {
   return dependencies;
 }
 
-function recurseRequires(scope: PackageScope) {
+function deepRequires(scope: PackageScope) {
   const deps = scope.dependencies || {};
   const reqs = Object.keys(scope.requires || {}).filter((req) =>
     deps[req] === undefined || !deps[req].bundled);
   for (const subscope of Object.values(deps)) {
     if (!subscope.bundled) {
-      reqs.push(...recurseRequires(subscope));
+      reqs.push(...deepRequires(subscope));
     }
   }
   return reqs;
--- a/cli/src/utils/getdeps.ts
+++ b/cli/src/utils/getdeps.ts
@@ -9,6 +9,12 @@ class MismatchedPackageAndPackageLockError extends CLIError {
   }
 }
 
+export interface PackageScope {
+  dependencies?: Record<string, PackageScope>;
+  requires?: Record<string, string>;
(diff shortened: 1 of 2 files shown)


cli/src/utils/getdeps.ts
```diff
@@ -16,20 +16,20 @@ export async function getDependencies(projectDir: string) {
   const { dependencies: lockDeps } = JSON.parse(await readFile(packageLockPath, 'utf8'));
   const { dependencies: pkgDeps } = JSON.parse(await readFile(packageJsonPath, 'utf8'));
 
-  const toProcess = Object.keys(pkgDeps);
+  const toProcess = Object.keys(pkgDeps).map((dep) => [dep, lockDeps]);
```
vogre: I think `root = lockDeps` is confusing - perhaps better to insert an empty object like `{}`. Also, this seems to break with npm dependencies https://www.npmjs.com/package/dependencies.

cli/src/utils/getdeps.ts
```diff
@@ -38,9 +38,23 @@ export async function getDependencies(projectDir: string) {
...
     // exclude already visited deps
     if (dependencies.has(p)) continue;
-    dependencies.add(p);
-    for (const subpackage of Object.keys(lockDeps[p].requires || {})) {
-      toProcess.push(subpackage);
+
+    const lockDep = lockDeps[p];
+    if (lockDep) {
+      dependencies.add(p);
+      if (has(root, 'dependencies')) {
```
vogre: Should it be `if (has(lockDep, 'dependencies'))` since root will never have dependencies?

bergundy:
@vogre I'll fix

cli/src/utils/getdeps.ts
```diff
@@ -38,10 +44,26 @@ export async function getDependencies(projectDir: string) {
...
+      console.error(`WARN: Cannot find dependency ${p} in package-lock.json, skipping upload`);
     }
   }
   return dependencies;
 }
+
+function recurseRequires(scope: PackageScope) {
+  const deps = scope.dependencies || {};
+  const reqs = Object.keys(scope.requires || {}).filter((req) =>
+    deps[req] === undefined || !deps[req].bundled);
```
bergundy: `node-pre-gyp` for example is bundled.

bergundy:
@vogre PTAnotherL when you get the chance, thanks!
Tested again on nextjs and marketplace CRM, both work after being deployed.

cli/src/utils/getdeps.ts
```diff
@@ -38,10 +44,26 @@ export async function getDependencies(projectDir: string) {
...
+      toProcess.push(...recurseRequires(lockDeps[p]));
+    } else {
+      // tslint:disable-next-line:no-console
+      console.error(`WARN: Cannot find dependency ${p} in package-lock.json, skipping upload`);
     }
   }
   return dependencies;
 }
+
+function recurseRequires(scope: PackageScope) {
```
vogre: Maybe call this `recurseDependencies` since we are recursing on the dependencies?

bergundy: I thought about it, but since it gets the attribute `requires` recursively I decided to reflect that in the name.
Not critical.

bergundy:
> I assume this was tested on the two examples.
Yes, it was tested.

cli/src/utils/getdeps.ts
bergundy: Renamed to `deepRequires`.

PR #320, opened 2019-11-17
Downgrade chalk package version

vogre:
Major version increase makes it incompatible with @oclif/color (not
currently causing any issues in our usage of color display, but confuses
the node_modules structure)

Reverts part of #318.


vogre:
The CI worked well enough - the behavior did not change, although it seems like the tests don't test the color output at all (seems like https://github.com/reshufflehq/reshuffle/blob/master/cli/src/test/snapshots/destroy.spec.ts.md has no terminal escape codes).

PR #321, opened 2019-11-18
React auth and integration

arielshaqed:
FE and major parts of BE by @bergundy.

Support Reshuffle OAuth.  Minimal unit tests for now, but an e2e template works.


passport-reshuffle/src/auth-handler.ts
```diff
@@ -0,0 +1,141 @@
...
+    });
+  })(req, res, next);
+}
+
+const oauthPage: express.Handler[] = [
+  passport.authenticate('auth0', { scope: 'openid email profile' }),
+  (_req, res) => res.redirect('/'),
+];
+
+// BUG: Use a remote shared per-app secret.
```
michaeladda: this is already the case. can remove th comment, no?

passport-reshuffle/src/auth-handler.ts
```diff
@@ -0,0 +1,141 @@
+import { makeStrategy, isFake } from './strategy';
+
+import express from 'express';
+import session from 'cookie-session';
+import passport from 'passport';
+import bodyParser from 'body-parser';
+
+const strategy = makeStrategy();
+
+passport.use(makeStrategy());
```
vogre: `passport.use(strategy)`?

passport-reshuffle/src/stick-figure.ts
```diff
@@ -0,0 +1,7 @@
+// A generic user found via a generic web search.  This is a
+// stick-figure of a person.
+// https://publicdomainvectors.org/en/free-clipart/Stick-Figure/58220.html,
+// labelled public domain, converted to JPG, scaled down and processed
```
michaeladda: uk for the win (`labelled`)

passport-reshuffle/src/strategy.ts
```diff
@@ -0,0 +1,76 @@
...
+import { Strategy as Auth0Strategy } from 'passport-auth0';
+import { titlecase } from 'stringcase';
+import { stickFigure } from './stick-figure';
+
+// Currently the *base* Passport Profile.  But could be anything
+// supported by the underlying Passport strategy.
+export { Profile };
+
+function getMissingOauthEnvariables() {
+  return ['OAUTH_DOMAIN', 'OAUTH_CLIENT_ID', 'OAUTH_CLIENT_SECRET']
```
vogre: My preference while we have control over those variables is to prefix them with `RESHUFFLE_`, since they are very close to what a user might use for their other, unrelated oauth https://github.com/reshufflehq/dev-docs/blob/master/backend/authBackend.js#L17.

passport-reshuffle/src/strategy.ts
```diff
@@ -0,0 +1,76 @@
...
+}
+
+class FakeLocalStrategy extends LocalStrategy {
+  public readonly fake: boolean = true;
+
+  // Just inherit the constructor from LocalStrategy.
+}
+
+export function makeStrategy(): Auth0Strategy | FakeLocalStrategy {
+  if (process.env.NODE_ENV === 'production' || process.env.OAUTH_CLIENT_ID) {
```
michaeladda: we should not decide if we're local or not basd on NODE_ENV, but this is not for this PR

passport-reshuffle/src/strategy.ts
```diff
@@ -0,0 +1,76 @@
...
+  _refreshToken: string,
+  _extraParams: any,
+  profile: Profile,
+  done: (error: any, user?: any, info?: any) => void,
+) {
+  // _extraParams.id_token has a JWT, if needed.
+  return done(null, profile);
+}
+
+function localVerifyUser(
```
michaeladda: fakeVerifyUser?
or use local elsewhre (local = fake iiuc)

passport-reshuffle/src/strategy.ts
michaeladda: what about `RESHUFFLE_SESSION_SECRET`?

passport-reshuffle/src/auth-handler.ts
```diff
@@ -0,0 +1,141 @@
...
+}
+
+const oauthPage: express.Handler[] = [
+  passport.authenticate('auth0', { scope: 'openid email profile' }),
+  (_req, res) => res.redirect('/'),
+];
+
+// BUG: Use a remote shared per-app secret.
+// TODO(ariels): Support secret rotation.
+const sessionSecretKey = process.env.SESSION_SECRET || 'fancy crab';
```
michaeladda: IIRC this is `RESHUFFLE_SESSION_SECRET` not `SESSION_SECRET`

react-auth/src/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+export const useAuthFlow = (): LoginManager => useContext(AuthContext).loginManager;
+
+// tslint:disable-next-line:variable-name
+export const AuthProvider: FC = ({ children }) => {
+  const { result, error, loading } = useAsync<AuthState>(async () => {
+    const req = await fetch('/whoami', { credentials: 'include' });
+    if (!req.ok) {
+      throw new Error(`Failed to get /whoami ${req.statusText}`);
+    }
+    // tslint:disable-next-line:no-return-await
```
vogre: OK, sorry, reversing my previous opinion here - less is more, and this makes the code consistent (reasoning https://github.com/reshufflehq/reshuffle/pull/176)

passport-reshuffle/src/auth-handler.ts
```diff
@@ -0,0 +1,141 @@
+import { makeStrategy, isFake } from './strategy';
```
michaeladda: I think the fake functions should live in a different file to make it clear they are only for the local dev use case
(not for this PR)

rush.json
```diff
@@ -423,6 +428,11 @@
         "packageName": "@reshuffle/utils-subprocess",
         "projectFolder": "utils-subprocess",
         "shouldPublish": true
+    },
+    {
+        "packageName": "@reshuffle/passport",
+        "projectFolder": "passport-reshuffle",
```
michaeladda: i think folder passport would have been better
(not for this PR)

passport-reshuffle/src/auth-handler.ts
arielshaqed: Ouch.  DOne.

passport-reshuffle/src/auth-handler.ts
arielshaqed: True.  Done.

passport-reshuffle/src/auth-handler.ts
arielshaqed: Yikes!

passport-reshuffle/src/strategy.ts
arielshaqed: Talk tomorrow.  I'm not sure which is better TBH, I'd like to poll some more people.

arielshaqed: It explicitly works without for local mode.

passport-reshuffle/src/strategy.ts
arielshaqed: Renamed.

passport-reshuffle/src/strategy.ts
arielshaqed: Ack.  But this *really* says that if you're in `production` you're not allowed to use fakes.  Or if you set one variable, you need to set them all.  Which I think is reasonable.

react-auth/src/index.tsx
arielshaqed: Well, I said there I'll do what the robot says, so here I'll do what the robot says.

arielshaqed:
Thanks!
I'll pull once tests pass.  @vogre -- you were not definitive on pull-or-not, so I'll fix anything needed tomorrow, of course.

PR #324, opened 2019-11-19
Use secure cookies

vogre:


vogre:
Tested and saw `secure: true` in browser dev tools.

PR #325, opened 2019-11-19
Proprly logout from OAUTH provider

michaeladda:
See https://auth0.com/docs/logout for more details

### DIFF
--- a/passport/src/auth-handler.ts
+++ b/passport/src/auth-handler.ts
@@ -4,6 +4,7 @@ import express from 'express';
 import session from 'cookie-session';
 import passport from 'passport';
 import bodyParser from 'body-parser';
+import { URLSearchParams } from 'url';
 
 const strategy = makeStrategy();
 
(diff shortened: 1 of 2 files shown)


passport/src/auth-handler.ts
```diff
@@ -140,7 +141,18 @@ export function authRouter(): express.Express {
 
   app.get('/logout', (req, res) => {
     req.logout();
-    res.redirect('/');
+    if (isFake(strategy)) {
```
arielshaqed: I'd rather keep the `isFake` call outside of the callback.

passport/src/auth-handler.ts
```diff
@@ -140,7 +141,18 @@ export function authRouter(): express.Express {
...
+      res.redirect('/');
+    } else {
+      const oauthDomain = process.env.OAUTH_DOMAIN!;
+      const baseUrl = process.env.RESHUFFLE_APPLICATION_DOMAINS!.split(',')[0];
+      const clientId = process.env.OAUTH_CLIENT_ID!;
+      const params = new URLSearchParams({
+        returnTo: `https://${baseUrl}`,
+        client_id: clientId,
+      }).toString();
+      res.redirect(`https://${oauthDomain}/v2/logout?${params}`);
```
arielshaqed: This also clears the session cookies?  Cool...

michaeladda: it doesn't clear it. it invalidates it afaict.

michaeladda:
passed in https://ci.binaris.io/job/gobinaris/job/precious/job/master/9623/ so merging

PR #326, opened 2019-11-19
Fix and share Profile interface

bergundy:


auth/src/index.ts
```diff
@@ -0,0 +1,12 @@
+export interface Email {
+  value: string;
+  type?: string;
+}
+
+export interface Profile {
```
vogre: It looks like the basics come from [passport.Profile](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/0e9f7eb3a2d623b0ce7d7b95ca8353c8b8e1b784/types/passport/index.d.ts#L138) extended by at least (`nickname`, `user_id`, `picture`) from [auth0](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/0e9f7eb3a2d623b0ce7d7b95ca8353c8b8e1b784/types/auth0/index.d.ts#L429) - can we create some union type out of them?

auth/src/index.ts
```diff
@@ -0,0 +1,12 @@
...
+  provider: string;
```
vogre: Looking at code and experimenting with it live, it looks like provider field is optional - see the profile [sources](https://github.com/auth0/passport-auth0/blob/master/lib/Profile.js#L4). Should we fill those or leave unchanged?

passport/package.json
```diff
@@ -27,6 +27,7 @@
   },
   "homepage": "https://github.com/reshufflehq/reshuffle#readme",
   "dependencies": {
+    "@reshuffle/auth": "0.1.0",
```
vogre: [Technicality]: I think this should be "0.0.0", for rush tracking to work?

react-auth/src/index.tsx
```diff
@@ -1,26 +1,8 @@
 import React, { FC, createContext, useContext } from 'react';
 import { useAsync } from 'react-async-hook';
+import { Profile } from '@reshuffle/auth';
```
vogre: I think `@reshuffle/react-auth` is quite generic, and it would work even if Profile looked like a generic `{}` object, and might be overridden by users - then we wouldn't need to export a package...

michaeladda:
FYI, if the API changes, this PR requires matching PRs for the Todo and template apps the least, as well as coordinating with @rylandg and @zolga

bergundy:
@michaeladda this fixes inconsistencies between the local and remote authentication strategies.

passport/package.json
bergundy: I'll fix

react-auth/src/index.tsx
bergundy: I'd rather give our users a consistent Profile interface.
In the future we might want to let users add more information to the basic profile.

auth/src/index.ts
bergundy: Let's leave it like this for now.

auth/src/index.ts
bergundy: IMHO It's easier for users to have a single consistent Profile interface so they don't have to guess which fields are in the profile.
We can extend the basic profile later.

bergundy:
@vogre PTAL

cli/src/test/browse.spec.ts
```diff
@@ -30,6 +31,7 @@ test('browse lists templates on dumb terminal', async (t) => {
       similarTemplatesIds: [
         'uniqueorn',
       ],
+      tags: [],
```
vogre: Is this from an unrelated change?

bergundy: Yes unfortunately

PR #327, opened 2019-11-20
Output CommonJS for react-auth

vogre:
Similar to react-async-hooks configuration
Required for Jest test runner


vogre:
This is actually not similar to react-async-hook configuration which uses [tsdx](https://github.com/jaredpalmer/tsdx) which has multiple outputs (esm, cjs.development, cjs.production).

vogre:
@bergundy PTAL - I've changed to tsdx, which seems to not output the generator stuff, and has the various formats.

PR #332, opened 2019-11-20
Export a pre-created authHandler

vogre:
To be similar in style with defaultHandler


vogre:
This is actually backward compatible, we can release a non-compatible version after.

PR #333, opened 2019-11-21
Clearly report backend misconfiguration

arielshaqed:
When the backend is misconfigured and `_handler` is missing the
default backend does not give a JSON response.  Fail accordingly and
print a readable error message.

This is end-user-visible, of course -- but so is the existing behavior of
printing
```
SyntaxError: Unexpected token < in JSON at position 0
```

### DIFF
--- a/common/changes/@reshuffle/react-auth/feature-better-_handler-error-message_2019-11-21-08-41.json
+++ b/common/changes/@reshuffle/react-auth/feature-better-_handler-error-message_2019-11-21-08-41.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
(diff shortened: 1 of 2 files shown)


arielshaqed:
New behaviour: this error message *in red*:
```
Error: Failed to GET /whoami JSON data: backend may be missing _handler.js.
```

arielshaqed:
Tested by packing and hiding `_handler.js` from `auth-template`.  Got desired error message.

arielshaqed:
Test failing due to webapp testing misconfiguration -- unrelated.  Pulling after consulting.
Thanks!

PR #334, opened 2019-11-21
Support auth flow in new window

bergundy:


passport/src/auth-handler.ts
```diff
@@ -123,6 +154,25 @@ export function createAuthHandler(): express.Express {
...
+      .end(`
+  <!doctype html>
+  <html lang="en">
+    <head>
+      <title>Redirect page</title>
+      <meta charset="utf-8">
+    </head>
+    <body>
+      <script>
+        window.localStorage.setItem('__reshuffle__login', new Date().toString());
```
arielshaqed: This is local storage, but auth is on a cookie.  Won't this cause weirdness?

arielshaqed:
# Thanks!

PR #336, opened 2019-11-21
More robust "Content-Type: application/json" test (this one works)

arielshaqed:


react-auth/src/index.tsx
```diff
@@ -47,8 +47,8 @@ export const AuthProvider: FC = ({ children }) => {
     if (!req.ok) {
       throw new Error(`Failed to get /whoami ${req.statusText}`);
     }
-    if (req.headers.get('content-type') !== 'application/json') {
-      throw new Error('Failed to GET /whoami JSON data: backend may be missing _handler.js.');
+    if (!req.headers.get('content-type')!.startsWith('application/json')) {
```
vogre: My "evil" server which does `res.end(JSON.stringify({}))` without `content-type` would crash this?

arielshaqed:
Thanks!

Pulling after tests.

PR #338, opened 2019-11-24
Use primary domain in the CLI

michaeladda:
Tested on my realm (which has multiple domains)

### DIFF
--- a/cli/src/commands/list.ts
+++ b/cli/src/commands/list.ts
@@ -2,6 +2,7 @@ import columnify from 'columnify';
 
 import Command from '../utils/command';
 import flags from '../utils/cli-flags';
+import { getPrimaryURL } from '../utils/helpers';
 
 export default class List extends Command {
   public static description = 'list applications';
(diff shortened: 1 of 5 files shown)


cli/src/utils/helpers.ts
```diff
@@ -52,3 +53,11 @@ export async function getProjectPackageJson() {
   }
   return loaded;
 }
+
+export function getPrimaryDomain(environment: Environment) {
+  return(`${environment.domains.filter((d) => d.type === 'subdomain')[0].name}`);
```
vogre: Are we relying on ordering or assuming exactly one subdomain? Could possibly be an additional property on a domain marking at as primary to avoid logic on client.

cli/src/commands/list.ts
```diff
@@ -40,7 +41,7 @@ export default class List extends Command {
           return;
         }
         this.log(columnify(mappedApps, {
-          columns: ['name', 'updatedAt', 'URL'],
+          columns: ['name', 'updatedAt', 'URL', 'customDomains'],
```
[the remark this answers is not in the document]
michaeladda: oops. leaked from future PR. removed.

cli/src/utils/helpers.ts
michaeladda: we still assume there's only one subdomain (the future PR does not assume order of cnames)

PR #339, opened 2019-11-26
Add server-side field for type-safely handling JWT auth token

arielshaqed:


interfaces/src/db.ts
```diff
@@ -116,6 +116,7 @@ export interface ClientContext {
 }
 
 export interface ServerOnlyContext {
+  verifiedToken: object | undefined; // Set on server from JWT
```
vogre: This can be written in the typescript syntax `verifiedToken?: object`, with the same JSON schema, but allowing omitting the property completely in a literal I think.

arielshaqed:
Thanks!
Integration won't pass before publishing the new version (and also had an unrelated integration flakiness failure), so pushing and force-publishing.

PR #340, opened 2019-11-26
Avoid type `object` in Concord objects

arielshaqed:
TJS makes it into `{}`, when we really want something more like the
typed replacement (which is good enough here).


interfaces/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@reshuffle/interfaces",
-  "version": "0.5.0",
+  "version": "0.5.1",
```
vogre: Need to replace "0.5.0" with "0.5.1" in all files due to rush<->concord hostility.

arielshaqed:
(Published 0.5.1 tagged as branch so integration will pass.  After I'll pull I'll re-tag as latest.)

arielshaqed:
`binaris-cli` flakiness is unrelated to this change; pulling.

Thanks!

PR #344, opened 2019-12-01
Support per-domain auth strategies

vogre:
Attempting to use the module without matching the domain name changes
the behavior to 404, since we don't fallback to a default domain - and
for security purposes probably shouldn't

### DIFF
--- a/passport/src/auth-handler.ts
+++ b/passport/src/auth-handler.ts
@@ -96,9 +96,11 @@ function createOnCallback(domain: string): express.Handler {
   return (req, res, next) => {
     passport.authenticate(`auth0-${domain}`, (err, user, _info, _extra) => {
       if (err) { return next(err); }
+      // TODO: redirecting to /login, a misconfiguration causes a confusing redirect loop
       if (!user) { return res.redirect('/login'); }
       req.logIn(user, (loginErr) => {
-        if (err) { return next(loginErr); }
+        // loginErr usually indicates a failure to serialize when using a session store
+        if (loginErr) { return next(loginErr); }
         res.redirect(req.session?.returnTo || '/');
         delete req.session?.returnTo;
       });
--- a/passport/src/auth-handler.ts
+++ b/passport/src/auth-handler.ts
@@ -1,4 +1,4 @@
-import { makeStrategy, isFake } from './strategy';
+import { makeStrategies, isFake } from './strategy';
 
 import express from 'express';
 import session from 'cookie-session';
@@ -6,9 +6,14 @@ import passport from 'passport';
 import bodyParser from 'body-parser';
 import { URLSearchParams } from 'url';
 
-const strategy = makeStrategy();
-
-passport.use(strategy);
+const strategies = makeStrategies();
+for (const { domain, strategy } of strategies) {
+  if (isFake(strategy)) {
+    passport.use(strategy);
+  } else {
+    passport.use(`auth0-${domain}`, strategy);
+  }
+}
 
 passport.serializeUser((user, cb) => cb(null, user));
 passport.deserializeUser((obj, cb) => cb(null, obj));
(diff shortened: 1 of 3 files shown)


passport/src/auth-handler.ts
```diff
@@ -1,14 +1,19 @@
...
-const strategy = makeStrategy();
-
-passport.use(strategy);
+const strategies = makeStrategies();
+for (const { domain, strategy } of strategies) {
+  if (isFake(strategy)) {
+    passport.use(strategy);
+  } else {
+    passport.use(`auth0-${domain}`, strategy);
+  }
```
arielshaqed: Consider:
```suggestion
passport.use(isFake(strategy) ? 'fake' : `auth0-domain`, strategy);
```

passport/src/auth-handler.ts
```diff
@@ -77,28 +82,58 @@ form {
...
+  };
+}
+
+function createOnCallback(domain: string): express.Handler {
+  return (req, res, next) => {
+    passport.authenticate(`auth0-${domain}`, (err, user, _info, _extra) => {
+      if (err) { return next(err); }
+      if (!user) { return res.redirect('/login'); }
+      req.logIn(user, (loginErr) => {
+        if (err) { return next(loginErr); }
```
arielshaqed: AFAICT this line is never called, because the containing function would have returned at line 98.  Should this be
```suggestion
        if (loginErr) { return next(loginErr); }
```
?

passport/src/auth-handler.ts
```diff
@@ -77,28 +82,58 @@ form {
...
+        res.redirect(req.session?.returnTo || '/');
+        delete req.session?.returnTo;
+      });
+    })(req, res, next);
+  }
+}
+
+function createOauthApp(domain: string) {
+  // using Router and not an appp to avoid duplicating app.set-s
```
arielshaqed: Maybe rename it `createOAuthRouter` (also note nitty caps)?

passport/src/auth-handler.ts
```diff
@@ -173,39 +210,35 @@ export function createAuthHandler(): express.Express {
...
+          res.redirect(req.session?.returnTo || '/');
+          delete req.session?.returnTo;
+        }
+      );
+    }
   }
 
-  if (!isFake(strategy)) {
-    app.get('/callback', onCallback);
+  for (const { domain, strategy } of strategies) {
```
arielshaqed: This may be easier to read if unified with the previous loop.

passport/src/strategy.ts
```diff
@@ -49,18 +49,29 @@ class FakeLocalStrategy extends LocalStrategy {
   // Just inherit the constructor from LocalStrategy.
 }
 
-export function makeStrategy(): Auth0Strategy | FakeLocalStrategy {
+
+export function makeStrategies() {
```
arielshaqed: I think this function belongs in `auth-handler`: it's emphatically *not* a strategy.

arielshaqed:
(Also "how was this tested", of course)

passport/src/auth-handler.ts
vogre: Good catch! Added comment - we are not expected to hit this line - since our `serializeUser` function can not throw, but still better to be correct.

passport/src/auth-handler.ts
```diff
@@ -96,9 +96,11 @@ function createOnCallback(domain: string): express.Handler {
   return (req, res, next) => {
     passport.authenticate(`auth0-${domain}`, (err, user, _info, _extra) => {
       if (err) { return next(err); }
+      // TODO: redirecting to /login, a misconfiguration causes a confusing redirect loop
       if (!user) { return res.redirect('/login'); }
       req.logIn(user, (loginErr) => {
-        if (err) { return next(loginErr); }
+        // loginErr usually indicates a failure to serialize when using a session store
```
arielshaqed: Yay for comments!  This one _really_ helps.

PR #345, opened 2019-12-01
Update packages

vogre:
Avoid using published @reshuffle/interfaces packages, and instead rely
on the generated files to ensure issues are spotted earlier

Any update to the interfaces version requires from all users to run
./scripts/generate_interfaces to updated @reshuffle/interfaces packages


vogre:
Already part of https://github.com/reshufflehq/reshuffle/commit/a03798d698dfb0d4e7a52592a602884a910febba#diff-b572c26f4ef17730e9097876b48ed68a.

PR #348, opened 2019-12-02
Show connected custom domain after deploy if exists

michaeladda:
manually tested by deploying both with and without custom domains

### DIFF
--- a/cli/src/utils/helpers.ts
+++ b/cli/src/utils/helpers.ts
@@ -55,7 +55,12 @@ export async function getProjectPackageJson() {
 }
 
 export function getPrimaryDomain(environment: Environment) {
-  return(`${environment.domains.filter((d) => d.type === 'subdomain')[0].name}`);
+  const { domains } = environment;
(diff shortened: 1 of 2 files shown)


cli/src/utils/helpers.ts
```diff
@@ -55,7 +55,12 @@ export async function getProjectPackageJson() {
 }
 
 export function getPrimaryDomain(environment: Environment) {
-  return(`${environment.domains.filter((d) => d.type === 'subdomain')[0].name}`);
+  const { domains } = environment;
+  const connectedCustomDomain = domains.find((d) => d.type === 'custom' && d.connected);
```
vogre: The `connected` check differs from the search in https://github.com/binaris/webapp/pull/219/files - should it be the same?

michaeladda: i don't think so.
my rational: here we give a link so it better work. in the webapp it's just the name in the menu, and when you click you see the url you click on
either way, as sen in this pr it is very easy to change :-)

PR #349, opened 2019-12-02
Allow configuring authentication token in test DB client

arielshaqed:


arielshaqed:
Force-pulling because rebasing over 81c084d..cc29085 brought no interesting changes.

PR #350, opened 2019-12-02
Improve lint

michaeladda:


cli/src/test/browse.spec.ts
```diff
@@ -20,7 +20,8 @@ test('browse lists templates on dumb terminal', async (t) => {
       description: 'something',
       similarTemplatesIds: [],
       tags: [],
-    }, {
+    },
```
arielshaqed: Why?  To my eye this just wastes space, and violates the spirit of this holy dictum https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace:
> The basic principle is: The more code that fits on one screen, the easier it is to follow and understand the control flow of the program. Use whitespace purposefully to provide separation in that flow.

I prefer to nestle commas here, if you don't have a strong preference otherwise.  If this linter configuration _requires_ it, I will argue here strongly for not requiring it.  I would prefer to force the nestled version here.

michaeladda: didn't try to start a war. simply enforcing a little more with minimal changes. removed elements alignment.

vogre:
Looks like it doesn't catch many cases - proposed https://github.com/reshufflehq/reshuffle/pull/352 instead.

PR #351, opened 2019-12-02
Allow configuring authentication token in test DB client (take 2)

arielshaqed:
*Real* Thoth tokens sign the application ID.  So token cannot be
 constant.  Instead pass a function that builds it.

### DIFF
--- a/common/changes/@reshuffle/db-testsuite/configure-thoth-auth-token_2019-12-02-16-08.json
+++ b/common/changes/@reshuffle/db-testsuite/configure-thoth-auth-token_2019-12-02-16-08.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@reshuffle/db-testsuite",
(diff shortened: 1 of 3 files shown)


common/changes/@reshuffle/db-testsuite/configure-thoth-auth-token_2019-12-02-16-08.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@reshuffle/db-testsuite",
+      "comment": "Use useful API to set auth token in tests: a real token requires signing the appId, support that.",
+      "type": "minor"
```
vogre: Note that need to upgrade to this version after.

arielshaqed: Thanks for making sure I don't forget!

PR #352, opened 2019-12-02
Eslint + whitespace rule

vogre:
Most significant changes:
- Type unions indented somewhat funny
- CallExpression always aligned with 2 spaces and not on first argument
- case statements indented on same level as switch

(for vim users: :set cinoptions=:0 to use this style)


cli/src/commands/logs.ts
```diff
@@ -15,29 +16,29 @@ export default class Logs extends Command {
...
-`// offset format
-$ ${Command.cliBinName} logs --since 3d
-$ ${Command.cliBinName} logs --since 13hours
-$ ${Command.cliBinName} logs --since 9s
-`,
-`// show all logs from 2 minutes ago and follow in real time
-$ ${Command.cliBinName} logs --since 2m --follow`,
-];
+    dedent`// retrieve all logs (except "function invocation")
+    $ ${Command.cliBinName} logs`,
```
arielshaqed: Really don't like this indentation, I'd rather indent this line further in.  (I'm guessing it will work, too!)

michaeladda:
I'm personally happy with any enforced style, and since this looks more comprehensive and uses eslint which we'll need to switch to anyhow, i'm in favor of this over #350

vogre:
Note that SwitchCase can be changed to previous style by using `"SwitchCase": 1` configuration, this is more of a preference.

arielshaqed:
> Note that SwitchCase can be changed to previous style by using `"SwitchCase": 1` configuration, this is more of a preference.

Actually, I photographed the page from the bible.  It says:
cover
how to indent switch

vogre:
Thanks!

Some possible improvements in future PRs:

- Make this configuration editor friendly (e.g. link to the eslintrc in `pacakge.json` or per folder `.eslintrc.js`)
- DRY between react/non-react configurations.

PR #353, opened 2019-12-04
Make deploy command CI friendly

bergundy:
Please do not review yet, this is not ready.


cli/src/commands/deploy.ts
```diff
@@ -172,14 +187,24 @@ export default class Deploy extends Command {
     return value === NEW_APPLICATION ? undefined : value;
   }
 
+  public async findApplicationIdByName(appName: string): Promise<string> {
```
michaeladda: CR: Use `const { id } = this.lycanClient.getAppByName(appName)`

cli/src/commands/deploy.ts
```diff
@@ -28,7 +31,7 @@ import {
 // * project associated deploys to associated app
 // * project associated and deploy fails shows meaningful message
 //
-// TODO: add --app-id and --env flags to bypass app association mechanism
+// TODO: add --env flag to bypass app association mechanism
```
michaeladda: CR: remove the `--env` TODO too

cli/src/test/deploy.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+    {
+      applicationId: 'abc',
+      defaultEnv: 'default',
+      directory: t.context.projectDir,
+    },
+  ]);
+});
+
+//
+// test('no args not in project dir', async (t) => {
```
michaeladda: why are these disabled?

cli/src/commands/deploy.ts
bergundy: To be able to deploy to the non `default` app environment, different from env variables.

cli/src/commands/deploy.ts
bergundy: Thanks!

cli/src/test/snapshots/deploy.spec.ts.md
```diff
@@ -0,0 +1,194 @@
...
+      out: `Building and bundling your app! This may take a few moments, please wait␊
+      ␊
+      > poll@ build PROJECT_DIR␊
+      > echo nothing to do␊
+      ␊
+      nothing to do␊
+      Preparing backend...␊
+      built␊
+      Uploading your assets! This may take a few moments, please wait␊
+      [?25l[36m?[39m [1mSelect target[22m [90m›[39m [90m- Use arrow-keys. Return to submit.[39m␊
```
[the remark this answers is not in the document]
bergundy: Good point, I can give it a try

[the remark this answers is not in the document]
bergundy: There doesn't seem to be an option in the `prompts` library to turn it off, internally it uses [kleur](https://www.npmjs.com/package/kleur) which supports turning it off but I'd rather not mess around with it for tests.
Filtering it manually doesn't seem ideal either.
Keeping it for now.

PR #354, opened 2019-12-04
Run tests on node12 and node10

vogre:
Update to use @types/node@12 
- cli
  Update to @types/mz@2.7.0 (more in sync with latest @types/node)
- utils-subprocess
  Workaround spawn types (use less specific overloads)
- local-proxy
  Use Buffer.from instead of deprecated `new Buffer`
  Type-assert that encoding is a valid BufferEncoding

Running integration for node10 in addition to node12 might slow down the process, we might want to drop support at some point, or optimize the verdaccio slowness.


utils-subprocess/src/index.ts
```diff
@@ -20,7 +20,11 @@ export async function spawn(
   options?: SpawnOptions
 ): Promise<void> {
   try {
-    await waitOnChild(origSpawn(command, args, options));
+    // Workaround @types/node - avoid choosing overloads per options.stdio variants
```
michaeladda: nice!

PR #355, opened 2019-12-05
Don't fail node_modules copy on missing deps

bergundy:


cli/src/utils/getdeps.ts
```diff
@@ -46,7 +50,7 @@ export async function getDependencies(projectDir: string) {
     if (dependencies.has(p)) continue;
 
     if (lockHasProperty(p)) {
-      dependencies.add(p);
+      dependencies.set(p, { optional: Boolean(lockDeps[p].optional) });
```
vogre: Boolean can be written as `!!`?

bergundy: I prefer `Boolean`.

bergundy:
@vogre exactly.

cli/src/utils/getdeps.ts
arielshaqed: You could always use the `[true][+...]` operator:
```js

> [true,false][+'']
true
> [true,false][+2]
undefined
> [true,false][+'s']
undefined
> [true][+'s']
undefined
> [true][+'']
true
> [true][+false]
true
> [true][+true]
undefined
> [true][+0]
true
> [true][+-2]
undefined
>
```

PR #358, opened 2019-12-09
Replace tslint with eslint

vogre:
According to npx tslint-to-eslint-config missing rules are
- strict-boolean-expressions
  changes in eslint to not allow unions marking a lot of valid uses
- jsdoc
  formatting of jsdoc comments

If needed, those can be implemented inside eslint-tslint plugin


common/.eslintrc.fast.json
```diff
@@ -2,22 +2,34 @@
...
   ],
   "extends": [
     "eslint:recommended",
     "plugin:@typescript-eslint/eslint-recommended",
     "plugin:@typescript-eslint/recommended"
   ],
   "rules": {
+    "@typescript-eslint/array-type": ["error", { "default": "array-simple", "readonly": "generic" }],
     "@typescript-eslint/ban-ts-ignore": "off",
+    "@typescript-eslint/consistent-type-definitions": "error",
```
arielshaqed: It's not clear from the [rule description](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/consistent-type-definitions.md) that a default type definition exists.  Can you specify explicitly `interface`?

common/.eslintrc.fast.json
```diff
@@ -2,22 +2,34 @@
...
+      {
+        "accessibility": "explicit",
+        "overrides": { "constructors": "no-public" }
+      }
+    ],
     "@typescript-eslint/explicit-function-return-type": "off",
     "@typescript-eslint/no-explicit-any": "off",
+    "@typescript-eslint/prefer-for-of": "error",
+    "@typescript-eslint/prefer-function-type": "error",
+    "@typescript-eslint/unified-signatures": "error",
```
arielshaqed: nit: not a fan.

common/.eslintrc.fast.json
```diff
@@ -45,6 +57,82 @@
     ],
     "lines-between-class-members": ["error", "always", { "exceptAfterSingleLine": true }],
     "indent": "off",
-    "@typescript-eslint/indent": ["error", 2, { "ignoreComments": true, "flatTernaryExpressions": true }]
+    "@typescript-eslint/indent": ["error", 2, { "ignoreComments": true, "flatTernaryExpressions": true }],
+    "guard-for-in": "error",
+    "new-parens": "error",
+    "no-bitwise": "error",
```
arielshaqed: I don't like this rule.  I think @bergundy doesn't, either.

Were it more nuanced I might change my mind.

common/.eslintrc.fast.json
```diff
@@ -45,6 +57,82 @@
...
+    "no-caller": "error",
```
arielshaqed: Hurrah!

common/.eslintrc.fast.json
```diff
@@ -45,6 +57,82 @@
...
+    "key-spacing": "error",
+    "space-before-blocks": "error",
+    "space-before-function-paren": "off",
+    "@typescript-eslint/space-before-function-paren": ["error", {
+        "anonymous": "never",
+        "named": "never",
+        "asyncArrow": "always"
+    }],
+    "space-in-parens": "error",
+    "space-unary-ops": "error",
```
arielshaqed: This makes sense (only) because of the default value.  Maybe specify it explicitly or in a comment (assuming this is JSON-with-comments), to be clear that this line is actually a _good_ thing?

common/.eslintrc.fast.json
```diff
@@ -45,6 +57,82 @@
...
+    "arrow-body-style": "error",
+    "arrow-parens": "error",
+    "dot-notation": "error",
+    "eol-last": "error",
+    "import/no-extraneous-dependencies": "error",
+    "max-len": [
+      "error",
+      {
+        "code": 120,
+        "ignoreRegExpLiterals": true
```
arielshaqed: Maybe also `"ignoreTemplateLiterals": true`?  You had to add a `+` above to a long template literal because of this.

passport/src/auth-handler.ts
```diff
@@ -164,7 +166,7 @@ function createPerDomainAuth(domain: string, strategy: passport.Strategy) {
     );
   }
   router.all('/logged-in', (_req, res) => {
-    return res.header('content-type', 'text/html')
+    res.header('content-type', 'text/html')
```
arielshaqed: Why did we have to get rid of `return` here?  React?

common/.eslintrc.fast.json
vogre: I guess it's most useful to catch non-short circuiting int32 `&` and `|`. It's possible to allow all others with `"allow": ["<<", "^", "~"...`, or disable the rule completely and be slightly more careful in codereviews? (Especially when using typescript I think danger is quite low).

common/.eslintrc.fast.json
vogre: It is the default in `tslint` to allow template literals, but I think for this use case (verbose error messages) they remain `grep`able even broken up on multiple lines.

common/.eslintrc.fast.json
vogre: I do like this rule in theory since overloads are not really a javascript feature - and a unified signature better represents the underlying function, but I do remember encountering false positives with this rule - can be omitted IMO.

cli/src/commands/deploy.ts
```diff
@@ -78,7 +78,7 @@ export default class Deploy extends Command {
           await mkdirp(target);
           await copy(source, target);
         } else if (!props.optional) {
-          // tslint:disable-next-line:no-console
+          /* tslint:disable-next-line:no-console*/ /* eslint-disable-next-line no-console */
```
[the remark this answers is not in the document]
vogre: Block comments always work. It's the C++ comments that work only for disable-next-line and not disable.

passport/src/auth-handler.ts
vogre: This is the same rule as the change in https://github.com/reshufflehq/reshuffle/pull/358/files#diff-3e34d3ded243ab6635707c51d43e10efR177 -- [arrow-body-style](https://eslint.org/docs/rules/arrow-body-style).

Not having a `return` allows to keep braces and indentation unchanged - it behaves slightly different than tslint `arrow-return-shorthand`.

PR #360, opened 2019-12-15
Implement storage backend

bergundy:
Tested while developing locally.
Will add tests using the e2e testsuite in a later PR once the rest of the pieces are in place.


storage/src/cloud.ts
```diff
@@ -0,0 +1,131 @@
...
+      Bucket: this.config.bucket,
+      Key: `tmp/${this.config.keyPrefix}/${objectId}`,
+      Expires: Math.floor(this.config.uploadExpirationMs / 1000),
+      ContentType: contentType,
+    });
+    return { token: objectId, signedUrl };
+  }
+
+  protected putOpts(objectId: string) {
+    const key = `${this.config.keyPrefix}/${objectId}`;
```
michaeladda: The id to key mapping is repeat several times. Please extract.

storage/src/storage.ts
```diff
@@ -23,6 +23,11 @@ if (process.env.NODE_ENV === 'production') {
     bucket: envStr('RESHUFFLE_STORAGE_S3_BUCKET'),
     keyPrefix: envStr('RESHUFFLE_STORAGE_S3_KEY_PREFIX'),
     uploadExpirationMs: ms('1h'),
+    credentials: {
+      accessKeyId: envStr('RESHUFFLE_STORAGE_AWS_ACCESS_KEY_ID'),
+      secretAccessKey: envStr('RESHUFFLE_STORAGE_AWS_SECRET_ACCESS_KEY'),
+      sessionToken: process.env.RESHUFFLE_STORAGE_AWS_SESSION_TOKEN,
```
michaeladda: why no envStr?

common/changes/@reshuffle/storage/storage-backend_2019-12-15-14-31.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@reshuffle/storage",
```
michaeladda: maybe call this user-content instead of storage?

storage/src/cloud.ts
```diff
@@ -0,0 +1,139 @@
...
+  protected key(objectId: string, prefix: string = ''): string {
+    return `${prefix}${this.config.keyPrefix}/${objectId}`;
+  }
+
+  public async createUpload(opts?: Partial<PutOptions>): Promise<{ token: string, signedUrl: string }> {
+    const objectId = uuid4();
+    const signedUrl = await this.s3.getSignedUrlPromise('putObject', {
+      Bucket: this.config.bucket,
+      Key: this.key(objectId, 'tmp/'),
+      Expires: Math.floor(this.config.uploadExpirationMs / 1000),
```
michaeladda: maybe ceil so you won't expire before expected? (expiring after is usually better accepted)

storage/src/handler.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  try {
+    res.json(await storage.createUpload({ contentType }));
+  } catch (err) {
+    logError(err);
+    res.sendStatus(500);
+  }
+});
+
+if (process.env.NODE_ENV !== 'production') {
```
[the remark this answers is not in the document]
michaeladda: ok. please document this.
also please extract the path to a const so it will be clear it's being reused elsewhere. it might not work if baseUrl != /storage for example.
also while /storage/upload/sign cannot conflict with /storage/upload/:id it is only since uuidv4 can't return the string sign. might want to change this too

storage/src/storage.ts
bergundy: It's optional

storage/src/cloud.ts
bergundy: Not critical IMHO.

storage/src/handler.ts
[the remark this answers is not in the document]
bergundy: `sign` and `:id` have different HTTP methods.
`baseUrl` is hardcoded here since there's no user exposed way of modifying it.
I'll document in a future PR.

PR #365, opened 2019-12-21
Add setDbOptions to allow overriding http options

vogre:


db-client/src/index.ts
```diff
@@ -41,9 +42,29 @@ const db = new DB(
...
+ * ## Example: increase timeouts for large queries
+ *
+ * ```js
+ * db.setDbOptions({ timeoutMs: 10000 });
+ * ```
+ */
+export function setDbOptions(options: Options = {}) {
+  if (options.timeoutMs !== undefined &&
+    (typeof options.timeoutMs !== 'number' || options.timeoutMs >= 60000 || options.timeoutMs <= 0)) {
+    throw new Error('Resshuffle DB configuration error: timeoutMs must be a number between 0 and 60000');
```
arielshaqed: Nit: "Reshuffle".

arielshaqed: Why an upper limit? 1 minute is plenty, but so are 30 seconds or 2 minutes.

arielshaqed:
Thanks!

db-client/src/index.ts
vogre: Thanks!

Added comment about the limit to hint about the suggested usage (we didn't have to limit, but it would be a surprise why queries are getting lost).

PR #366, opened 2019-12-21
Print validation errors in debug mode

vogre:


vogre:
> I wonder if we should log this too somewhere like .reshuffle/logs/cli.log similar to .reshuffle/logs/reshuffle.log we use for the local server.

We probably should, opened a task for it. We should probably highlight logs in "new issue template" for those to be visible.

PR #370, opened 2019-12-25
Add --new-app flag

vogre:
To support non-interactive use from CI


cli/src/commands/deploy.ts
```diff
@@ -194,7 +198,10 @@ export default class Deploy extends Command {
   }
 
   public async run() {
-    const { flags: { 'app-name': givenAppName, env: givenEnv } } = this.parse(Deploy);
+    const { flags: { 'app-name': givenAppName, env: givenEnv, 'new-app': forceNewApp } } = this.parse(Deploy);
+    if (givenAppName && forceNewApp) {
+      this.error('--app-name and --new-app flags are incompatible');
```
arielshaqed: Not sure why, actually?

cli/src/commands/deploy.ts
```diff
@@ -220,10 +227,10 @@ export default class Deploy extends Command {
 
     this.startStage('register app on local');
     const env = 'default'; // hardcoded for now
-    let project = findProjectByDirectory(projects, projectDir);
+    let project = forceNewApp ? undefined : findProjectByDirectory(projects, projectDir);
 
     let application: Application;
-    if (!project || givenAppName) {
+    if (!forceNewApp && (!project || givenAppName)) {
```
arielshaqed: This logic is getting _very_ hairy.  In particular, `forceNewApp` only ever appears in the negative... which makes for interesting reading.

At a micro level, I *think* this conditional might be easier to read in its negated dual form:
```suggestion
    if (!(forceNewApp || (project && !givenAppName)() {
```

cli/src/test/deploy.spec.ts
```diff
@@ -120,6 +120,15 @@ test('given --app-name but app not in list gives informative message', async (t)
   t.snapshot({ ...result, out: result.out.replace(t.context.projectDir, 'PROJECT_DIR') });
 });
 
+test('project associated deploys to new app with --new-app', async (t) => {
+  const newApp = createApp({ id: 'hybrid-koala' }); // hybrid-koala != fluffy-samaritan (default)
+  td.when(t.context.uploadFake()).thenReturn({ ok: true, digest: 'abcabc' });
+  td.when(t.context.lycanFake.deployInitial(anything, 'default', 'abcabc', [])).thenResolve(newApp);
+  const result = await t.context.shell.run(`${t.context.run} deploy --new-app`, 'utf-8');
+  t.snapshot({ ...result, out: result.out.replace(t.context.projectDir, 'PROJECT_DIR') });
```
arielshaqed: Not sure this entirely tests  the desired "new app" behaviour.

cli/src/test/deploy.spec.ts
```diff
@@ -120,6 +120,15 @@ test('given --app-name but app not in list gives informative message', async (t)
   t.snapshot({ ...result, out: result.out.replace(t.context.projectDir, 'PROJECT_DIR') });
 });
 
+test('project associated deploys to new app with --new-app', async (t) => {
+  const newApp = createApp({ id: 'hybrid-koala' }); // hybrid-koala != fluffy-samaritan (default)
```
arielshaqed: What would happen if you passed `fluffy-samaritan`?  Should that also be a test??

cli/src/commands/deploy.ts
```diff
@@ -194,7 +198,10 @@ export default class Deploy extends Command {
   }
 
   public async run() {
-    const { flags: { 'app-name': givenAppName, env: givenEnv } } = this.parse(Deploy);
+    const { flags: { 'app-name': givenAppName, env: givenEnv, 'new-app': forceNewApp } } = this.parse(Deploy);
```
arielshaqed: It seems OClif can do this.  Perhaps use `exclusive` (https://oclif.io/docs/flags)?

cli/src/commands/deploy.ts
```diff
@@ -218,40 +225,43 @@ export default class Deploy extends Command {
...
-          directory: projectDir,
-          applicationId,
-          defaultEnv: env,
-        };
-        // Don't update association if given --app-name option
-        if (!givenAppName) {
-          projects.push(project);
-          this.conf.set('projects', projects);
-        }
+        this.startStage('register app on local');
```
arielshaqed: 1. This times local execution.
2. Having it conditional will make analytics harder.  Let's either not have a stage, or just `this.startStage('start')` before starting.

arikmaor:
> To support non-interactive use from CI

We need to have an API key support for the CI...

vogre:
> We need to have an API key support for the CI...

We currently do something like 
```bash
export RESHUFFLE_CONFIG=/path/to/reshuffle.config.yml
echo accessToken: $RESHUFFLE_ACCESS_TOKEN > $RESHUFFLE_CONFIG
```

PR #375, opened 2020-01-05
Ensure runningOn element exists before calling then

vogre:
This fixes a race condition with HTML loading before react rendering


vogre:
From running again, looks like this is the wrong fix.

vogre:
@michaeladda PTAL - changed so that both branches have assertions, upgraded to Cypress 3.8.1 resolved the hanging on `"hack"` test.

PR #377, opened 2020-01-06
Extract build function to seperate package

arikmaor:


arikmaor:
I understand why this build breaks.
What I don't understand, is why does it pass on `master` :\
@michaeladda @vogre Do you have a clue?

vogre:
@arikmaor: the way things currently work is that the concord packages build almost by accident because they don't depend on typescript. You need to make sure `pqckage.json` dependencies are aligned (carets, `@types/node@12`, etc) with other packages and then `typescript` is globally available.

vogre:
You can reproduce locally by temporarily `npm remove -g typescript`

build-utils/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "test": "echo No tests yet"
+  },
+  "files": [
+    "/dist"
+  ],
+  "keywords": [],
+  "author": "",
+  "license": "MIT",
+  "dependencies": {
+    "@reshuffle/utils-subprocess": "~0.0.2",
```
vogre: Nit^extreme: can we haz carets?

build-utils/src/index.ts
```diff
@@ -0,0 +1,149 @@
...
+  logger: console,
+};
+
+function escapeWin32(filePath: string) {
+  return process.platform === 'win32' ? shellEcape(filePath) : filePath;
+}
+
+export async function build(projectDir: string, options?: Partial<BuildOptions>): Promise<string> {
+  const {
+    skipNpmInstall,
```
vogre: Nit: this is already new code and not moved old code

build-utils/src/index.ts
```diff
@@ -0,0 +1,149 @@
...
+      'Content-Type': 'application/gzip',
+      'Content-Length': `${contentLength}`,
+    },
+    body: stream,
+  });
+  const json = await res.json();
+  if (res.status !== 200) {
+    // TODO: check error if response is not a json and display a nice error message
+    const { message } = await res.json();
+    logger.error(message);
```
vogre: Due to comment in L145 we might never reach this line, but when we do, previous behavior was `oclif.error` which logs and then throws a `CLIError` and just logs.

build-utils/src/index.ts
```diff
@@ -0,0 +1,149 @@
...
+      ...apiHeaders,
+      'Content-Type': 'application/gzip',
+      'Content-Length': `${contentLength}`,
+    },
+    body: stream,
+  });
+  const json = await res.json();
+  if (res.status !== 200) {
+    // TODO: check error if response is not a json and display a nice error message
+    const { message } = await res.json();
```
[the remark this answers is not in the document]
arikmaor: fixed

build-utils/src/index.ts
arikmaor: correct
is this a problem?

build-utils/src/index.ts
arikmaor: I've handled this scenario but catching the error and re-throwing the `CLIError` in the `cli` project
https://github.com/reshufflehq/reshuffle/pull/377/files#diff-e88bfc89b1330c5aed4c81329ea15882R99

build-utils/package.json
arikmaor: changed
`rush update` didn't create changes, is that alright?

build-utils/src/index.ts
```diff
@@ -0,0 +1,149 @@
...
+  }, ['.']);
+  return tarPath;
+}
+
+export async function uploadCode(
+  tarPath: string,
+  url: string,
+  apiHeaders: Record<string, string> = {},
+  logger = DEFAULT_OPTIONS.logger
+) {
```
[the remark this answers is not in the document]
arikmaor: I'll do it in a future PR
I think this function might move back to the CLI
Uploading to `spice` is a bit different then uploading to `S3`
it practically becomes a wrapper around `fetch` without abstracting much more

build-utils/package.json
vogre: I think it's OK.

Also for `@reshuffle/utils-subprocess` specifically should be exact `"0.0.2"` since rush will manage the exact dependency when upgrading.

build-utils/src/index.ts
vogre: I think this is a different kind of error - `logger.error` can be replaced with a ```throw new Error(`failed to upload: ${message}`);``` or similar.

Example of error response status is 500 and body is `{"message": "entity.too.large"}`.

common/changes/@reshuffle/build-utils/feature-add-build-utils_2020-01-07-14-33.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "comment": "",
+      "packageName": "@reshuffle/build-utils",
+      "type": "none"
```
vogre: Might be worthwile to mark this change as `minor` so it is published with a `0.1.0` version so we can have semi-semver for it.

PR #378, opened 2020-01-07
Allow backend/ dir to be missing

vogre:


vogre:
Thanks!

Tested with a `create-react-app my-app` and seeing it deployed without adding a dummy backend directory.

> Can we extract the string `backend` to somewhere else? I reckon we'll have to allow _some_ configuration of this, but maybe not in the staging area, so we _really_ need to be careful about creating more copies of it.

Good idea, will send a future PR.

> If I _don't_ have a `backend` directory and build a bad frontend project, do I get different error diagnostics than if I _do_ have such a directory but no `@expose`d functions?

The `code-transform/macro` will refuse to work with imports from a `server/` (any non `backend/` directory). It's also possible to create a `backend/` directory with only a `_handler.js` but no exposed functions.

PR #380, opened 2020-01-09
Fix parsing of port argument

vogre:


vogre:
https://github.com/reshufflehq/reshuffle/commit/c986a5f0509793883530facd5418290fb47a3953 fixes for commander syntax (missing `<port>`).

Yargs is not strictly better, but at least it will be consistent and familiar.

PR #383, opened 2020-01-13
Hide values in "reshuffle env --list", and quote them on display

arielshaqed:
1. By default `reshuffle env --list` shows all values as `<hidden>`.
2. To include values, `reshuffle env --list --include-values`.
3. Both `--list` or `--get` quote values, allowing a single
   interpretation for `<hidden>`, `undefined`, and `"value"`.


cli/src/commands/env.ts
```diff
@@ -100,11 +118,15 @@ export default class Env extends Command {
     }
   }
 
-  protected async list(appId: string) {
-    const { variables } = await this.lycanClient.getEnv(appId, null);
-    // Output in sorted order (of variable names).
-    const sorted = variables.map(vvToString).sort();
-    this.log(sorted.join('\n'));
+  protected async list(appId: string, includeValues: boolean) {
+    try {
```
vogre: No need for `try-catch` - I think errors are handled at a different level.

cli/src/commands/env.ts
```diff
@@ -100,11 +118,15 @@ export default class Env extends Command {
...
+      const { variables } = await this.lycanClient.getEnv(appId, null);
+      // Output in sorted order (of variable names).
+      const sorted = (includeValues ? variables : eraseValues(variables)).map(vvToString).sort();
```
vogre: `sort` without a comparator function is suspicious, although a `=` separator guarantees that we are sorting by variable names and not the whole value.

cli/src/commands/env.ts
arielshaqed: Yuck, that's debugging code.  Squashing it out at the source (and removing this stain on code history).

cli/src/commands/env.ts
arielshaqed: Indeed, the output string itself is sorted.

arielshaqed:
Yup, I know about these weirdo snaps.

Thanks!  Publishing this once tests pass...

PR #384, opened 2020-01-14
Include values in "env --list" by default rather than hide them

arielshaqed:
Change the name of the flag from `--include-values` to
`--hide-values`, with the relevant default.  Details on
[Slack](https://binarisio.slack.com/archives/C4QUEPS9G/p1578959693003700?thread_ts=1578929085.003200&cid=C4QUEPS9G).


cli/src/commands/env.ts
```diff
@@ -73,9 +73,9 @@ export default class Env extends Command {
       description: 'list all variables',
       default: false,
     }),
-    'include-values': flags.boolean({
-      char: 'i',
-      description: 'include values of all variables with --list',
+    'hide-values': flags.boolean({
+      char: 'x',
```
vogre: Is there a mnemonic for `x`?

arielshaqed: "x out those values" is the best I could do.  `-h` doesn't work because I expect it to be `--help`.

arielshaqed:
Thanks!

PR #385, opened 2020-01-14
Serialized deploy result

bluTomer:


cli/src/commands/deploy.ts
```diff
@@ -54,6 +54,11 @@ export default class Deploy extends Command {
                           Will create an app with a random name, for renaming use $ ${Command.cliBinName} rename`,
       exclusive: ['app-name'],
     }),
+    json: flags.boolean({
```
vogre: [list command](https://github.com/reshufflehq/reshuffle/blob/96ab87bb4ff2c94b57ee79d35c9210812ba1a8a1/cli/src/commands/list.ts#L18) uses a `--format` flag - let's use the same format for consistency.

cli/src/commands/deploy.ts
```diff
@@ -115,8 +120,15 @@ export default class Deploy extends Command {
     return app.id;
   }
 
+  public log(str: string) {
+    const { flags: { json } } = this.parse(Deploy);
```
vogre: This is a bit strange, since we are parsing the command line arguments on each run - I think no reason for not using a member `quiet: boolean` and then writing

```
if (this.quiet) return;
super.log(str);
```

cli/src/commands/deploy.ts
```diff
@@ -115,8 +120,15 @@ export default class Deploy extends Command {
...
+    if (!json) {
+      super.log(str);
```
vogre: `log` method is defined as [log(message?: string, ...args: any[])](https://github.com/oclif/command/blob/6be030b420f17c861823ba68f3f0d74cedd3852a/src/command.ts#L132) - we never used the extra arguments, but probably best to keep supporting them.

cli/src/commands/deploy.ts
```diff
@@ -191,6 +204,11 @@ export default class Deploy extends Command {
       const { applicationId, defaultEnv } = project;
       application = await this.lycanClient.deploy(applicationId, defaultEnv, digest, envVars);
     }
-    this.log(`Project successfully deployed! Your project is now available at: ${makeAppLink(application)}`);
+    const jsonApp = { ...application, link: makeAppLink(application) };
+    this.log(`Project successfully deployed! Your project is now available at: ${jsonApp.link}`);
+
+    if (json) {
+      this.log(JSON.stringify(jsonApp));
```
vogre: This wouldn't write anything to the screen (since we overwritten log).

cli/src/commands/deploy.ts
bluTomer: Will do

cli/src/commands/deploy.ts
bluTomer: I thought about this and I agree, not sure how to scope the new member though.

cli/src/commands/deploy.ts
bluTomer: Will do, was basing the method signature according to the [Logger interface](https://github.com/reshufflehq/reshuffle/blob/96ab87bb4ff2c94b57ee79d35c9210812ba1a8a1/build-utils/src/index.ts#L12) but using the interface for oclif's log is the more correct way

cli/src/commands/deploy.ts
bluTomer: fixed

cli/src/commands/deploy.ts
bluTomer: fixed

cli/src/commands/deploy.ts
bluTomer: done

build-utils/src/index.ts
```diff
@@ -32,27 +34,30 @@ export async function build(projectDir: string, options?: Partial<BuildOptions>)
...
+    quiet,
   } = {
     ...DEFAULT_OPTIONS,
     ...options,
   };
   logger.log('Building and bundling your app! This may take a few moments, please wait');
   const stagingDir = await mkdtemp(pathResolve(tmpdir(), 'reshuffle-bundle-'), { encoding: 'utf8' });
 
+  const stdioInherit = [0, 1, 2];
+  const stdioSuppress = [0, 2, 2];
```
vogre: Note that this is not exactly `suppress` (which I would assume is `redirect to /dev/null`, but `redirectToStderr`).

Both are probably useful in some cases, but the naming should be consistent (since this is not exactly `quiet`).

Perhaps our command line can have an additional `--quiet` flag which suppresses output, and does not even print it to stderr.

cli/src/commands/deploy.ts
```diff
@@ -115,8 +123,15 @@ export default class Deploy extends Command {
     return app.id;
   }
 
+  public log(message?: string, ...args: any[]) {
```
vogre: (not necessarily this PR): I wonder if it makes sense to move this to the `BaseCommand` class`.

cli/src/commands/deploy.ts
```diff
@@ -191,6 +207,11 @@ export default class Deploy extends Command {
       const { applicationId, defaultEnv } = project;
       application = await this.lycanClient.deploy(applicationId, defaultEnv, digest, envVars);
     }
-    this.log(`Project successfully deployed! Your project is now available at: ${makeAppLink(application)}`);
+    const jsonApp = { ...application, link: makeAppLink(application) };
```
vogre: Looking at `makeAppLink`, it uses `terminalLink` whicch on iterm2 behaves in a funny way - it produces escape codes so the link is clickable.

It does not appear when piping into `jq`, but it's still less useful - I think for the json output it would make sense to just call to `getPrimaryURL`, also maybe rename `link` to `primaryURL` (since there is sometimes more than one).

bluTomer: Done

cli/src/test/deploy.spec.ts
```diff
@@ -153,3 +153,20 @@ test('deploy fails with --app-name and --new-app', async (t) => {
...
+  t.context.projectConfig = JSON.stringify({ accessToken: 'setec-astronomy' });
+  await writeFile(t.context.configPath, t.context.projectConfig);
+
+  // ReJSONing app object to match the date format for the resulting 'createdAt' and 'updatedAt'.
+  const jsonApp = JSON.stringify(app);
+  const appWithLink = { ...JSON.parse(jsonApp), link: 'https://a.b.c' };
+  const jsonResult = await t.context.shell.run(`${t.context.run} deploy --format json`, 'utf-8');
+  let result;
+  t.notThrows(() => {result = JSON.parse(jsonResult.out);});
+  t.deepEqual(appWithLink, result);
```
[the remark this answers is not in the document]
bluTomer: Test now uses snapshot.

cli/src/commands/deploy.ts
```diff
@@ -191,6 +208,12 @@ export default class Deploy extends Command {
       const { applicationId, defaultEnv } = project;
       application = await this.lycanClient.deploy(applicationId, defaultEnv, digest, envVars);
     }
+
     this.log(`Project successfully deployed! Your project is now available at: ${makeAppLink(application)}`);
+
+    if (format === 'json') {
```
vogre: Could be
```suggestion
    if (this.jsonFormat) {
```

cli/src/test/deploy.spec.ts
```diff
@@ -153,3 +153,15 @@ test('deploy fails with --app-name and --new-app', async (t) => {
...
+test('deploy with json format returns only app json', async (t) => {
+  const app = createApp({ id: 'abc' });
+  td.when(t.context.lycanFake.listApps(anything)).thenResolve([]);
+  td.when(t.context.uploadFake()).thenReturn({ ok: true, digest: 'abcabc' });
+  td.when(t.context.lycanFake.deployInitial(anything, 'default', 'abcabc', [])).thenResolve(app);
+  t.context.projectConfig = JSON.stringify({ accessToken: 'setec-astronomy' });
+  await writeFile(t.context.configPath, t.context.projectConfig);
+
+  const jsonResult = await t.context.shell.run(`${t.context.run} deploy --format json`, 'utf-8');
+  t.snapshot(JSON.parse(jsonResult.out));
```
vogre: Can you snapshot `t.snapshot(result)` - so that both `result.out` and `result.err` have snapshots?

build-utils/src/index.ts
```diff
@@ -17,11 +17,13 @@ export interface Logger {
 export interface BuildOptions {
   skipNpmInstall: boolean;
   logger: Logger;
+  outputToErr: boolean;
```
arikmaor: I think this will be a bit more flexible
```suggestion
  spawnOptions?: SpawnOptions,
```
we already know of a third option we for stdio (in the builder-function context)

You can import `SpawnOptions` from `child_process`

You can use "object spread" for this:
``` js
x = { a: 1, b: 2 }
y = {
  a: 3,
  c: 5,
  ...x,
}
// y => { a: 1, c: 5, b: 2 }
```

build-utils/src/index.ts
```diff
@@ -6,6 +6,7 @@ import tar from 'tar';
 import shellEcape from 'any-shell-escape';
 import { spawn } from '@reshuffle/utils-subprocess';
 import { getDependencies, MismatchedPackageAndPackageLockError } from './getdeps';
+import { SpawnOptions } from 'mz/child_process';
```
arikmaor: The spawn used here, gets this interface from `child_process`, not `mz/child_process`
please change it for safety

cli/src/commands/deploy.ts
```diff
@@ -129,11 +145,17 @@ export default class Deploy extends Command {
     // TODO: select target before this block
     this.startStage('build');
 
+    const stdioInherit = [0, 1, 2];
```
arikmaor: nit: This is the default, so if `format === logs`,  you can just avoid passing `spawnOptions` all together (Not a blocker for this PR)

PR #387, opened 2020-01-15
Add DB iterator

arielshaqed:
Initial version to gauge usefulness of the interface.  In particular,
this does _not_ view a snapshot -- items added during iteration might
appear (or they might not).

### DIFF
--- a/common/changes/@reshuffle/labs-db-iterator/db-iterator_2020-01-15-13-52.json
+++ b/common/changes/@reshuffle/labs-db-iterator/db-iterator_2020-01-15-13-52.json
@@ -3,7 +3,7 @@
     {
       "packageName": "@reshuffle/labs-db-iterator",
       "comment": "New package: DB iterator",
-      "type": "patch"
+      "type": "minor"
     }
   ],
   "packageName": "@reshuffle/labs-db-iterator",
--- a/common/changes/@reshuffle/labs-db-iterator/db-iterator_2020-01-15-13-52.json
+++ b/common/changes/@reshuffle/labs-db-iterator/db-iterator_2020-01-15-13-52.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@reshuffle/labs-db-iterator",
+      "comment": "New package: DB iterator",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@reshuffle/labs-db-iterator",
+  "email": "ariels@reshuffle.com"
+}
--- a/labs/db-iterator/package.json
+++ b/labs/db-iterator/package.json
@@ -1,0 +1,44 @@
+{
+  "name": "@reshuffle/labs-db-iterator",
+  "version": "0.0.1",
+  "description": "Iterator for large DB queries",
+  "main": "dist/index.js",
+  "scripts": {
+    "build": "tsc",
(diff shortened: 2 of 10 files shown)


arielshaqed:
@avnerbraverman @rylandg I think you two wanted iteration over long DB queries.

labs/db-iterator/src/iterator.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+    db,
+    dbDir,
+  };
+});
+
+test.afterEach(async (t) => {
+  await rmrf(t.context.dbDir);
+});
+
+async function iteratorToArray<T>(it: AsyncIterableIterator<T>): Promise<T[]> {
```
vogre: It might makes sense to export this (trivial) function too. Or in documentation suggest using https://www.npmjs.com/package/streaming-iterables for utilities like `collect` or `reduce`.

common/changes/@reshuffle/labs-db-iterator/db-iterator_2020-01-15-13-52.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@reshuffle/labs-db-iterator",
+      "comment": "New package: DB iterator",
+      "type": "patch"
```
vogre: minor?

labs/db-iterator/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "@reshuffle/labs-db-iterator",
+  "version": "0.0.1",
+  "description": "Iterator for large DB queries",
+  "main": "dist/index.js",
+  "scripts": {
+    "build": "tsc",
+    "lint": "eslint -c ../../common/.eslintrc.json --ext .ts src",
+    "test": "RESHUFFLE_DB_BASE_URL=a RESHUFFLE_APPLICATION_ID=b RESHUFFLE_APPLICATION_ENV=c RESHUFFLE_ACCESS_TOKEN=d ava"
```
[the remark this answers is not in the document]
arielshaqed: Yuck, thanks!  Done.
Unfortunately we do need this because we always load the client.  I think I'll switch the default DB client to do "lazy" loading of environment variables via an access function; that way if you don't use the default client you won't need these variables.  That will go in a later PR.

labs/db-iterator/src/iterator.spec.ts
arielshaqed: Doing the latter.  I don't want to provide utilities to simplify this, it's an antipattern when using an async iterable.  It makes sense in this test, but is less useful generally.

Also you're right: we should gauge interest in different interfaces by providing `db-stream` and maybe even `db-observable`.

labs/db-iterator/src/iterator.ts
```diff
@@ -0,0 +1,27 @@
+import { find, Q } from '@reshuffle/db';
+import { DB } from '@reshuffle/db/dist/db';
+import { Document } from '@reshuffle/interfaces-node-client/interfaces';
+
+export interface IteratorProps {
+  chunkSize: number;
+}
+
+const defaultIteratorProps: IteratorProps = {
+  chunkSize: 100000,
```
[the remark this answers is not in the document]
arielshaqed: None.  And it's probably too high, so slicing it to 500.

PR #388, opened 2020-01-15
Redirect to /login-error on login error

vogre:


passport/README.md
```diff
@@ -49,3 +49,7 @@ export async function getTodos() {
   return get(`/todos/${user.id}`);
 });
 ```
+
+### Error handling
+
+If a login fails due to some reason (e.g. misconfiguration of the login provider) the user will be redirected to /login-error.
```
arielshaqed: Please consider adding an example of something useful to do there.

passport/src/auth-handler.ts
```diff
@@ -102,8 +102,7 @@ function createOnCallback(domain: string): express.Handler {
   return (req, res, next) => {
     passport.authenticate(`auth0-${domain}`, (err, user, _info, _extra) => {
       if (err) { return next(err); }
-      // TODO: redirecting to /login, a misconfiguration causes a confusing redirect loop
-      if (!user) { return res.redirect('/login'); }
+      if (!user) { return res.redirect('/login-error'); }
```
arielshaqed: How does this change the behaviour of existing apps, specifically of current templates with auth and any apps built on top of those?

vogre: I will submit PRs to update the package in all first-party templates.

passport/README.md
vogre: Good idea, added.

PR #389, opened 2020-01-15
Add advanced configuration section

vogre:


passport/README.md
```diff
@@ -49,3 +49,18 @@ export async function getTodos() {
   return get(`/todos/${user.id}`);
 });
 ```
+
+### Advanced Configuration
+
+When using with the reshuffle platform an OAuth provider will be configured. If
```
arielshaqed:
```suggestion
The Reshuffle platform automatically configures an OAuth provider. If
```

passport/README.md
```diff
@@ -49,3 +49,18 @@ export async function getTodos() {
...
+you wish to use an alternative oauth implementation you can override the
```
arielshaqed:
```suggestion
you wish to use an alternative OAuth implementation, override the
```

passport/README.md
```diff
@@ -49,3 +49,18 @@ export async function getTodos() {
...
+following environment variables:
+
+- `OAUTH_DOMAIN` \
+  Domain of the OAuth provider (e.g. `oauth.example`)
```
arielshaqed: Not sure this is a good example, it doesn't look like a DNS domain (and, indeed, AFAICT `example.` is not a TLD).

Maybe just give example values for Google or GitHub or some equally popular provider?  Or use `*.example.com`, which _is_ registered by IANA for use in examples.

vogre: `example` is a TLD, but it really doesn't look like a TLD...

arielshaqed: # Spec fight!  :-) 

https://www.icann.org/resources/pages/tlds-2012-02-25-en doesn't show .example.. Neither does whois.

vogre: https://www.iana.org/assignments/special-use-domain-names/special-use-domain-names.xhtml shows `example.` along with `example.com.`, `example.net.` and `example.org`. Yo are right about whois, it doesn't show anything.

PR #390, opened 2020-01-15
Keep backend dir structure in local proxy

michaeladda:
Tested by running `npm start` with binaris/marketplace-cms#21

### DIFF
--- a/local-proxy/src/server.ts
+++ b/local-proxy/src/server.ts
@@ -1,5 +1,5 @@
 import { promisify } from 'util';
-import { resolve as resolvePath, extname } from 'path';
+import { resolve as resolvePath, extname, relative } from 'path';
 import { Handler as DBHandler } from '@reshuffle/leveldb-server';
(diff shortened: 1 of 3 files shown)


local-proxy/src/server.ts
```diff
@@ -126,6 +127,12 @@ async function transpileAndCopy() {
       return extname(src) !== '.js';
     },
   });
+  try {
+    // Do we need to copy additional files from the root path?
```
vogre: I think this deserves thorough explanation to what is going on.

This was first raised by: https://github.com/reshufflehq/reshuffle/pull/243#discussion_r326894282

When we moved from `.reshuffle_local_proxy` to `.reshuffle/local_proxy`.

I think we can reuse code from https://github.com/reshufflehq/reshuffle/blob/d72c85a368193112257aed75842ac69768fa3aa0/build-utils/src/index.ts#L72 to copy everything but `backend`, `node_modules`, `src`, to have as close as possible emulation of the remote environment - e.g. some other code might want to access the `public/` directory.

PR #395, opened 2020-08-26
Ref1

lagache:


examples/hello-world-http/registerUnregisterHTTPExample.js
```diff
@@ -0,0 +1,21 @@
+const {HttpService, Reshuffle} = require('../../');
+
+const app = new Reshuffle();
+const service = new HttpService();
+
+app.register(service);
+
+// Check 2 starts
+app.start();
+app.start();
```
ashevat, line 10: why do we need to two app.start()?

index.js
```diff
@@ -1,127 +1,115 @@
...
-    }
+    event.getService = this.getService.bind(this);
+    
+    eventHandlers.forEach(handler => {
+      handler.handle(event);
+    });
+   
     return true;
   }
-  _p_handle(handler, event) {
```
ashevat, line 111: this is important to keep it separated for enterprise to overwrite how handle() is managed.

index.js
```diff
@@ -1,127 +1,115 @@
...
-  }
   registerHTTPDelegate(path, delegate) {
-    if (!this.shareResources.webserver) {
-      var express = require('express');
-      this.shareResources.webserver = express();
-      this.shareResources.webserver.route("*")
-        .all((function (req, res, next) {
-          let handled = false;
-          if (this.httpDelegetes[req.url]) {
-            handled = this.httpDelegetes[req.url].handle(req, res, next);
```
ashevat, line 46: getting the true/false from the delegate is important, the edge case is that a script is removed from a path - the delegate is still there and no one returns an HTTP response on the socket.

lib/http.js
```diff
@@ -37,43 +36,32 @@ class HttpService {
...
-      if(eventConfiguration.options.path === path &&  eventConfiguration.options.method === method){
-        console.log("HTTP got called, firing event");
-        handled = this.app.handleEvent(eventConfiguration.id, { "req": req, "res": res });
-      }
-    }
+    const {method, url} = req;
+    const eventConfiguration = Object.values(this.eventConfigurations)
+        .find(({options}) => options.path === url && options.method === method)
 
-    if (!handled){
```
ashevat, line 59: We need to return true false, or else the http call hangs if no scripts are connected.

examples/hello-world-http/registerUnregisterHTTPExample.js
lagache: oh it was part of my test to make sure the app doesn't crash and doesn't try to create a second express web server. This file will go once we have sufficient test coverage

lib/http.js
lagache: okay, I'll revert those

index.js
lagache: okay, I'll revert those

index.js
lagache: okay I'll revert this one, it will make sense to me one day I'm sure :)

PR #398, opened 2020-08-27
Datastore initial implementation 

ashevat:


examples/hello-world-persistent/helloWorldSQLExample.js
```diff
@@ -0,0 +1,30 @@
...
+  console.log(`Hello World! ${times} times.`);
+  times++;
+  await store.set("scripts/times-said-hello", times);
+
+  // for destributed setup with many reshuffle servers set() should be replaced with update()
+  let safe_count_times = await store.get('scripts/times-said-hello-safe_count_times') || 0;
+  console.log(`Hello World! safe count = ${safe_count_times} times.`);
+  await store.update("scripts/times-said-hello-safe_count_times", (safe_count_times) => {
+    return safe_count_times === undefined ? 1 : (safe_count_times+1);
+  });
```
avnerbraverman, line 26: No need to read from the database twice. You can simply do:

await store.update('scripts/times-said-hello-safe_count_times', (safeCountTimes = 0) => {
  console.log(`Hello World! safe count = ${safeCountTimes} times.`);
  return safeCountTimes + 1;
}

lib/index.js
```diff
@@ -2,4 +2,5 @@ module.exports = {
     CronService: require('./cron').CronService,
     HttpService: require('./http').HttpService,
     SlackService: require('./slack').SlackService,
+    SQLStoreStategy: require('./sqlStoreStategy').SQLStoreStategy
```
avnerbraverman, line 5: The name strategy is confusing. Maybe "backend"?

lib/sqlStoreStategy.js
```diff
@@ -0,0 +1,107 @@
...
+    await this.pool.query(
+      `CREATE TABLE IF NOT EXISTS ${this.table} ` +
+      '(id varchar primary key, value varchar);'
+    );
+    this.dbValidated = true;
+    return true;
+  }
+
+  validateKey(key) {
+    if (key && key.length === 0) {
```
avnerbraverman, line 21: This means that "undefined" and "0" are actually valid keys. If we want to adhere to my doc then you need something like:

if (typeof key !== 'string' || key.length === 0) {
  throw new Error(`Datastore: Invalid key: ${key}`);
}

(adding the key to the exception helps later when debugging)

lib/sqlStoreStategy.js
```diff
@@ -0,0 +1,107 @@
...
+    this.validateKey(key);
+
+   
+    const conn = await this.pool.connect();
+
+    try {
+      await conn.query('BEGIN');
+
+      const res = await conn.query(
+        `SELECT value FROM ${this.table} WHERE id = $1 FOR UPDATE`,
```
avnerbraverman, line 70: I found "FOR UPDATE" in a PostgreSQL article. If we want to support AnyDB we need to verify that this works in any SQL database. Nimo might know.

examples/hello-world-persistent/helloWorldSQLExample.js
ashevat: your implementation is too hard to understand in an example, I think.

lib/index.js
ashevat: what does backend mean? everything here is backend @avnerbraverman. I took Strategy from npm Passport.

lib/sqlStoreStategy.js
ashevat: good point

lib/sqlStoreStategy.js
ashevat: I we can revisit this later

duckranger: Generally speaking this will work with Postgres, MySQL (Innodb anyway), Oracle, DB2... 
The only strange behaviour will be with (of course) SQL Server - but only in edge cases.

duckranger (approved):
Danke

PR #399, opened 2020-08-27
Setup typescript

lagache:
files in the `dist` folder are unbearable but generated so don't look at them

Examples are not changed and still working as usual.
There's a mix of js and ts, we can refactor when we want, it works with both

linting is into place


duckranger (approved):
Great Work!

PR #402, opened 2020-08-31
refactored to connectors

ashevat:


lagache (approved):

examples/hello-world-http/registerUnregisterHTTPExample.js
```diff
@@ -1,21 +1,21 @@
-const {HttpService, Reshuffle} = require('../..');
+const { HttpConnector, Reshuffle} = require('../..');
```
duckranger, line 1: Do you think in all our examples we should consistently import `Reshuffle` first? So 
```suggestion
const { Reshuffle, HttpConnector} = require('../..');
```

duckranger (approved):
Verified by running build again (getting rid of `dist` and greping for the word service

examples/hello-world-http/registerUnregisterHTTPExample.js
lagache: I think we should format code in examples. I'll change it to do so and also add a rule to sort imports out so it's consistent.

PR #404, opened 2020-09-01
Use reshuffle base connector

lagache:


examples/hello-world-http/HTTPConnectorActionExamples.js
```diff
@@ -1,14 +1,17 @@
 const {CronConnector, HttpConnector, Reshuffle} = require('../..');
 
+const httpConnectionName = 'myHttpConnection'
+
 const app = new Reshuffle();
+
 const cronConnector = new CronConnector();
-const httpConnector = new HttpConnector();
+const httpConnector = new HttpConnector(undefined, httpConnectionName);
```
ashevat, line 8: maybe more common to pass null rather than undefined

PR #408, opened 2020-09-02
Fix the if on start

duckranger:
Http connector had the wrong check


duckranger:
Will be superseded by PR 409

PR #415, opened 2020-09-03
Updates to docs

duckranger:
Added smtp
Updated build connector tutorial

### DIFF
--- a/docs/building-connectors.md
+++ b/docs/building-connectors.md
@@ -112,7 +112,7 @@ app.when(monitor.on({type:'server-down'},'server-crashed'), (event) => {
 app.start();
 ```
 
-When your code calls `monitor.on({type:"server-down"},"server-crashed")`, the connector registers the EventConfiguration provided. It registers the handler subscribing to this specific event, and returns the EventConfiguration object:
+When your code calls `monitor.on({type:"server-down"},"server-crashed")`, the connector registers the EventConfiguration provided. Reshuffle then registers the handler subscribing to this specific event.
 ````js
 on(options, eventId) {
   if (!eventId) {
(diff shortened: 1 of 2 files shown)


docs/building-connectors.md
```diff
@@ -1,114 +1,136 @@
...
-class MyConnector{
+class MyConnector {
     
-constructor(connectorOptions, id) {...}
+  constructor(options, id) {...}
+
+  onStart(app) {...}
 
-start(app) {...}
+  onStop() {...}
```
ashevat, line 23: I think need to have both the stop and start and the onstop and on start. You need to implement the former if you are not inheriting from base connector.  Especially in light of the paragraph above.

docs/connectors.md
```diff
@@ -6,7 +6,7 @@ Reshuffle achieves this by using **Connectors** as the connection mechanism to a
 Reshuffle ships with an existing ecosystem comprising several connectors you can use to connect to common systems. A list of these Reshuffle-provided connectors can be found in this page, and their code can be found in the various repositories [here](https://github.com/reshufflehq/).  
 
 The framework is extendable, meaning you can create connectors to services you use or even your own proprietary systems. 
-The process for building new connectors is described in the [manual](./building-connectors.md). 
+The process for building new connectors is described in the [manual]("./buidling-connectors.md").
```
ashevat, line 9: The link here is broken, the removed link works and tested.

docs/building-connectors.md
duckranger: Good point. If you don't use `BaseConnector` then you need start and stop and if you do use it you need `onStart` and `onStop`. What would be a clear way to explain this?

docs/connectors.md
```diff
@@ -6,7 +6,7 @@ Reshuffle achieves this by using **Connectors** as the connection mechanism to a
 Reshuffle ships with an existing ecosystem comprising several connectors you can use to connect to common systems. A list of these Reshuffle-provided connectors can be found in this page, and their code can be found in the various repositories [here](https://github.com/reshufflehq/).  
 
 The framework is extendable, meaning you can create connectors to services you use or even your own proprietary systems. 
-The process for building new connectors is described in the [manual](./building-connectors.md). 
+The process for building new connectors is described in the [manual](./buidling-connectors.md).
```
ashevat, line 9: this link it still broken. seems like a merge issue. it should be [manual](./building-connectors.md)

ashevat (changes requested):

ashevat (approved):

PR #416, opened 2020-09-03
Use event.context with HttpConnector 

lagache:
Fix HTTPConnector to call handle with an event that has system attributes under context


duckranger (approved):

duckranger (approved):
Thanks

PR #417, opened 2020-09-04
defaulted persistent store to memory and removed persistent wrapper

ashevat:


src/Reshuffle.ts
```diff
@@ -136,13 +136,16 @@ export default class Reshuffle {
...
-    return ps
+    this.registry.common.persistentStore = adapter;
+    return adapter;
   }
 
   getPersistentStore() {
-    return this.registry.common.persistentStore
+
+    if(this.registry.common.persistentStore){
+      this.registry.common.persistentStore = new MemoryStoreAdapter();
```
duckranger: Maybe use the setter here? then you can do
`return this.registry.persistentStore || setPersistentStore(new MemoryStoreAdapter());`

duckranger (approved):
Look at the comment?

PR #418, opened 2020-09-04
Add ability to log to files and filter logs

lagache:


src/Logger.ts
```diff
@@ -0,0 +1,68 @@
...
+    warn: console.warn,
+    error: console.error,
+    debug: console.debug,
+  }
+
+  constructor(options = loggerOptions) {
+    this.logger = createLogger(options)
+  }
+
+  redirectConsoleToLogger(): void {
```
ashevat, line 47: I do not want to highjack console in Reshuffle fabric. as a library it is an anti pattern.

src/Reshuffle.ts
```diff
@@ -18,12 +19,21 @@ export default class Reshuffle {
...
+  logger: Logger
 
-  constructor() {
+  constructor(useLogger = true) {
     this.availableConnectors = availableConnectors
     this.port = parseInt(<string>process.env.PORT, 10) || 8000
     this.httpDelegates = {}
     this.registry = { connectors: {}, handlers: {}, common: {} }
+    this.logger = new Logger()
+    if (useLogger) {
```
ashevat, line 30: this should be moved to enterprise.

ashevat (changes requested):

src/Logger.ts
lagache: removed

src/Reshuffle.ts
lagache: Amir and I decided on keeping this at the end as Watson is a standard log package and allow the user to filter what he wants to be displayed in the console/log files

src/Logger.ts
```diff
@@ -0,0 +1,45 @@
+import { createLogger as winstonCreateLogger, format, transports } from 'winston'
+const { colorize, combine, printf, json, timestamp } = format
+
+const consoleFormat = printf(({ level, message, timestamp, scriptId }) => {
+  return `${timestamp} ${scriptId ? 'script' : 'runtime'} ${level} ${message}`
+})
+
+const filterLog = (withScriptId: boolean) =>
+  format((info) => (info.scriptId ? (withScriptId ? info : false) : withScriptId ? false : info))
```
duckranger: Can do something like `(typeof info.scriptId === 'undefined') == withScriptId ? info : false`

src/Logger.ts
```diff
@@ -0,0 +1,45 @@
...
+    //
+    // - Write all logs with level `error` and below to `error.log`
+    // - Write all logs with level `info` and below to `combined.log`
+    //
+    new transports.File({ filename: './logs/error.log', level: 'error' }),
+    new transports.File({ filename: './logs/combined.log' }),
+    new transports.File({ filename: './logs/script.log', format: filterLog(true)() }),
+    new transports.Console({
+      format: combine(
+        // filterLog(false)(),
```
duckranger, line 25: Can we remove commented out code?

src/Logger.ts
```diff
@@ -0,0 +1,45 @@
...
+  format((info) => (info.scriptId ? (withScriptId ? info : false) : withScriptId ? false : info))
+
+const loggerOptions = {
+  level: 'info',
+  format: combine(timestamp(), json()),
+  exitOnError: false,
+  transports: [
+    //
+    // - Write all logs with level `error` and below to `error.log`
+    // - Write all logs with level `info` and below to `combined.log`
```
duckranger: What does below mean in this context?

package.json
```diff
@@ -35,7 +35,8 @@
     "nanoid": "^3.1.12",
     "node-fetch": "^2.6.0",
     "prettier": "^2.1.1",
-    "reshuffle-base-connector": "0.0.1"
+    "reshuffle-base-connector": "0.0.1",
```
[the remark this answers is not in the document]
lagache: yes, I'll do it in a separate PR

src/Logger.ts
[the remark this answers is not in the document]
lagache: I pasted this from Winston examples but yeah I agree, doesn't make sense.
I'll change it

src/Logger.ts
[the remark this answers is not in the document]
lagache: it's for keeping the code temporary for the enterprise as we'll need to push to a different file for script logs.

src/Logger.ts
lagache: I'll remove this function, it will be in the enterprise but surely we can simplify the logic

src/Logger.ts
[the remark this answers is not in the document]
lagache: I'll keep it aside and remove it from this PR

duckranger (approved):
Great thanks!

PR #422, opened 2020-09-07
Separate options for on vs constructor

lagache:


src/connectors/CronConnector.ts
```diff
@@ -2,21 +2,21 @@ import { BaseConnector, EventConfiguration } from 'reshuffle-base-connector'
...
-    super(options, id)
+  constructor(id?: string) {
+    super(undefined, id)
     this.intervalsByEventId = {}
   }
 
-  on(options: CronConnectorOptions, eventId: string) {
+  on(options: CronEventOptions = DEFAULT_EVENT_OPTIONS, eventId: string) {
     if (!eventId) {
       eventId = `CRON/${options.interval}/${this.id}`
```
duckranger, line 21: That's not going to be unique if no id is passed to the constructor

duckranger (approved):
Approving because it is possibly not a huge issue but -
I think it is likely that a timer will have 2 events with the same interval, and the way the code is currently - the second event will override the first one.

src/connectors/CronConnector.ts
lagache: It's actually the opposite, it might not be unique if you pass the same id.
But if you don't pass an id, BaseConnector uses `nanoid` to generate one so it will be unique

PR #423, opened 2020-09-08
Connector on method returns an EventConfiguration

lagache:
3 TS error because of `on` signature different from baseConnector class


duckranger (approved):
Thanks!!

PR #430, opened 2020-09-11
new cron-connector

gabeeden:


examples/cron/SimpleCronExample.js
```diff
@@ -0,0 +1,15 @@
+const { Reshuffle, CronConnector } = require('../..')
+const app = new Reshuffle()
+
+const connector = new CronConnector(app)
+
+const logger = app.getLogger()
+
+// eslint-disable-next-line no-unused-vars
+connector.on({ expression: '*/1 * * * * *' }, (event) => {
```
lagache, line 9: I think we should still allow passing interval. It could be one or the other as the cron expression is not intuitive.

options could be either {interval: '5'} or { expression: '*/5 * * * * *' }
@duckranger @ashevat @gabeeden what do you think?

lagache:
I think `setInterval` and `clearInterval` in Reshuffle.ts aren't used and can be removed.
You'll also need to remove it from 'interface Reshuffle' in `reshuffle-base-connector`

examples/cron/SimpleCronExample.js
gabeeden: I think we are going to have a separate Timer connector

duckranger: I'd rather have a single connector and allow more robust options. e.g. the `CronEventOptions` could have a `frequency: MIN | HOUR | DAY | WEEK ...` and the `on` method can translate these to the right expression. 
Could also add 'startAt' and `finishAfter` if we wanted to add more power.

src/connectors/CronConnector.ts
```diff
@@ -1,63 +1,49 @@
...
 
-  on(
-    options: CronEventOptions = DEFAULT_EVENT_OPTIONS,
-    handler: any,
-    eventId?: string,
-  ): EventConfiguration {
+  on(options: CronEventOptions, handler: any, eventId?: string): EventConfiguration {
     if (!eventId) {
-      eventId = `CRON/${options.interval}/${this.id}`
+      eventId = `CRON/${this.id}`
```
duckranger, line 20: This sets up for not creating a unique id. I would urge to add a `nanoid()` or something

src/connectors/CronConnector.ts
```diff
@@ -1,63 +1,49 @@
...
-        this.app?.handleEvent(event.id, event)
-      }, event.options.interval)
-      this.intervalsByEventId[event.id] = intervalId
-    }
+    const task = cron.schedule(
+      event.options.expression,
+      () => this.app.handleEvent(event.id, event),
+      { timezone: event.options.timezone, scheduled: false },
+    )
+    this.tasksByEventId[event.id] = task
```
duckranger, line 33: Together with the comment above: If the user doesn't supply an event id - this here will remove previous tasks

duckranger (approved):
Have a look at the event-id comments

src/connectors/CronConnector.ts
gabeeden: Yes true, will add that

duckranger (approved):
Thanks

PR #444, opened 2020-09-27
Slack example

lagache:


examples/message/SlackMessageExample.js
```diff
@@ -0,0 +1,94 @@
+const { Reshuffle } = require('reshuffle')
+const { SlackConnector, SlackEventType, SlackEvents } = require('reshuffle-slack-connector')
+const app = new Reshuffle()
+
+const main = async () => {
+  const slackConnector = new SlackConnector(app, {
```
ashevat, line 6: can you add a link to where to get these tokens from?

examples/message/SlackMessageExample.js
```diff
@@ -0,0 +1,94 @@
...
+    token: '<your_slack_token_starting_with_xox>',
+    signingSecret: '<your_slack_signing_secret>',
+    port: '<slack_receiver_port>', // Default to 3000
+    endpoints: '<slack_receiver_endpoints>', // Default to '/'
+  })
+
+  // Events
```
ashevat, line 13: probably needs documentation on how to set these events up in Slack

examples/message/SlackMessageExample.js
```diff
@@ -0,0 +1,94 @@
...
+  slackConnector.on(
+    {
+      type: SlackEventType.EVENT,
```
ashevat, line 16: SlackEventType.EVENT sounds confusing

examples/message/SlackMessageExample.js
lagache: added

examples/message/SlackMessageExample.js
lagache: added

examples/message/SlackMessageExample.js
lagache: I kept the same semantic as runtime v1. I don't have a better name to suggest but I think it makes sense when you look at the other types

ashevat (approved):

PR #451, opened 2020-11-03
Rs 144 http delegates prioritisation

lagache:


src/Reshuffle.ts
```diff
@@ -105,8 +104,22 @@ export default class Reshuffle {
...
+    if (Object.keys(this.httpDelegates).length) {
+      const webserver = this.prepareWebServer()
+      Object.keys(this.httpDelegates)
+        .sort()
+        .reverse() // The sort().reverse() moves all generic routes (/:id) at the end
+        .forEach((path) => {
+          const httpMultiplexer = this.httpDelegates[path]
+          webserver.all(path, httpMultiplexer.handle.bind(httpMultiplexer))
+        })
+      webserver.all('*', (req, res) => {
```
lagache, line 116: This is the default route after everything, it means it wasn't handled before so we return  501

src/Reshuffle.ts
```diff
@@ -175,14 +188,20 @@ class HttpMultiplexer {
...
+    if (this.delegates.length > 0) {
+      for (const delegate of this.delegates) {
+        if (!handled) {
+          handled = await delegate.handle(req, res, next)
+        }
+      }
+    }
+
+    if (!handled) {
+      next()
```
lagache, line 204: if not handled, go find the next route (e.g. /:id )

src/Reshuffle.ts
```diff
@@ -175,14 +188,20 @@ class HttpMultiplexer {
...
     }
-    this.delegates.forEach(async function (delegate) {
```
lagache, line 184: Something I learnt the hard way, the forEach is asynchronous, a for loop with await is synchronous!

test/Reshuffle.spec.ts
```diff
@@ -305,8 +305,8 @@ describe('Reshuffle', () => {
         app.start()
 
         const responseTest3Call = await request(app.registry.common.webserver).get('/unknownPath')
-        expect(responseTest3Call.status).toBe(404)
-        expect(responseTest3Call.text).toContain('Cannot GET /unknownPath')
+        expect(responseTest3Call.status).toBe(501)
```
lagache, line 308: That's the only unit tests I couldn't do properly before, and now all make sense 🥇

src/Reshuffle.ts
```diff
@@ -105,8 +104,22 @@ export default class Reshuffle {
     Object.values(this.registry.connectors).forEach((connector) => connector.start())
 
     // Start the webserver if we have http delegates
-    if (this.registry.common.webserver) {
-      this.httpServer = this.registry.common.webserver.listen(this.port, () => {
+    if (Object.keys(this.httpDelegates).length) {
+      const webserver = this.prepareWebServer()
```
duckranger, line 108: I like how now we only have a webserver if there are HTTP delegates 👍

src/Reshuffle.ts
```diff
@@ -175,14 +188,20 @@ class HttpMultiplexer {
...
+  async handle(req: any, res: any, next: any) {
     req.originalPath = this.originalPath
-    if (this.delegates.length === 0) {
-      const errorMessage = `No handler registered for ${req.method} ${req.originalPath}`
-      return res.status(501).send(errorMessage)
+    let handled = false
+
+    if (this.delegates.length > 0) {
+      for (const delegate of this.delegates) {
+        if (!handled) {
```
duckranger, line 197: Maybe we can do `if (handled) break;` ?

duckranger (approved):
Thanks :)

src/Reshuffle.ts
lagache: yes, good idea

duckranger (approved):

src/Reshuffle.ts
lagache: can't register new routes after start() but I don't think it ever works

PR #455, opened 2020-11-09
Tidy delegate registration

duckranger:


src/Reshuffle.ts
```diff
@@ -63,10 +63,8 @@ export default class Reshuffle {
     let httpMultiplexer = this.httpDelegates[path]
```
lagache, line 63:
```suggestion
    const httpMultiplexer = this.httpDelegates[path] || new HttpMultiplexer(path)
    httpMultiplexer.delegates.push(delegate)
```

ashevat (approved):

src/Reshuffle.ts
```diff
@@ -60,13 +60,12 @@ export default class Reshuffle {
   }
 
   registerHTTPDelegate(path: string, delegate: BaseHttpConnector): Reshuffle {
-    let httpMultiplexer = this.httpDelegates[path]
+    const httpMultiplexer = this.httpDelegates[path] || new HttpMultiplexer(path)
```
ashevat, line 63: If this is const then the assignment in a few lines fails.

ashevat (changes requested):

src/Reshuffle.ts
```diff
@@ -60,13 +60,12 @@ export default class Reshuffle {
...
+    httpMultiplexer.delegates.push(delegate)
     if (!httpMultiplexer) {
```
lagache, line 65: This if needs to be deleted as it is done line 63

src/Reshuffle.ts
lagache: line 65 and 67 needs to be deleted, catered by the `|| new HttpMultiplexer(path)`, and so the const will remain ;)

ashevat: @lagache  see build error:
src/Reshuffle.ts:66:7 - error TS2588: Cannot assign to 'httpMultiplexer' because it is a constant.

lagache: @ashevat yes I know, sorry I made one mistake in my comment above, line 65 **TO** 67 need to be removed.
the if is replaced by the `|| new HttpMultiplexer(path)`
@duckranger do you confirm?

lagache (approved):

PR #456, opened 2020-11-09
RS-47 server health check

lagache:


src/Reshuffle.ts
```diff
@@ -113,6 +115,9 @@ export default class Reshuffle {
           const httpMultiplexer = this.httpDelegates[path]
           webserver.all(path, httpMultiplexer.handle.bind(httpMultiplexer))
         })
+      webserver.use(HEALTH_CHECK_PATH, (req, res) =>
```
ashevat, line 118: is there a way to tell Reshuffle that you do not want a health check?

lagache: done, only make health check available when HEALTH_CHECK_PATH is set.
It's covered by the unit tests

ashevat (approved):

PR #462, opened 2020-11-24
RS-174 - Shopify Example

lagache:


examples/shopify/shopify.js
```diff
@@ -0,0 +1,41 @@
...
+// 2. Click `Create new private app`,
+// 3. Provide a name and scopes (keeping in mind that permissions are very limited per default)
+// 4. Click Save, and you'll get api key and password
+
+const app = new Reshuffle()
+const connector = new ShopifyConnector(app, {
+  shopName: '<your_shop_name>',
+  apiKey: '<your_shopify_private_app_api_key>',
+  password: '<your_shopify_private_app_password>',
+  baseURL: '<your_runtime_base_url>',
```
duckranger, line 18: Might be nice to use `process.env` here?

examples/shopify/shopify.js
```diff
@@ -0,0 +1,41 @@
+const { Reshuffle } = require('reshuffle')
+const { ShopifyConnector } = require('reshuffle-shopify-connector')
+
+// This example uses reshuffle-shopify-connector
+// Code and documentation available on Github: https://github.com/reshufflehq/reshuffle-shopify-connector
+
```
duckranger, line 6: Maybe add what the example does? e.g.
1. Lists up to 5 open orders
2. Lists all products
3. Creates a new country

duckranger (approved):
Thanks :) Look at comments please to see if they make sense

PR #467, opened 2020-12-03
Add specific webhook message

duckranger:


src/Reshuffle.ts
```diff
@@ -113,6 +113,11 @@ export default class Reshuffle {
           res.status(200).send({ ok: true, uptime: process.uptime() }),
         )
       }
+      webserver.all('/webhooks/*', (req, res) => {
```
ashevat, line 116: why do we need this? Seems very use case specific and not general use.. maybe move to enterprise?

lagache: it's a good backup mechanism. If the runtime receive webhook events and it's not implemented then we reply with something meaningful to the devs

lagache (approved):

PR #470, opened 2020-12-10
RS-190 - custom event

zacmos:


examples/customEvent/CustomEventExample.js
```diff
@@ -0,0 +1,20 @@
+const { CustomEventConnector, Reshuffle } = require('../..')
+
+const app = new Reshuffle()
+const connector = new CustomEventConnector(app)
+
+connector.on({ channel: 'One', payLoad: { name: 'Jack', age: 25 } }, (event) => {
```
ashevat, line 6: channel seems like a new concept here, why not use something like `event` or 'eventTriggered'

src/connectors/CustomEventConnector.ts
```diff
@@ -0,0 +1,31 @@
...
+  on(
+    options: CustomEventConnectorEventOptions,
+    handler: any,
+    eventId?: string,
+  ): EventConfiguration {
+    if (!eventId) {
+      eventId = `CustomEvent/${options.channel}/${this.id}`
+    }
+    const event = new EventConfiguration(eventId, this, options)
+    this.eventConfigurations[event.id] = event
```
ashevat, line 21: This means that you old event handler for the same string will be overrun... is that the desired behavior?

ashevat (dismissed):

examples/customEvent/CustomEventExample.js
duckranger: I think channel is to allow event grouping (like topic/queue name)

lagache, line 6: payload instead of payLoad

lagache (changes requested):

src/connectors/CustomEventConnector.ts
zacmos: As far as I know this is how the connectors are implemented, for the same eventId new event will be set.
I updated the fire() function to handle case when user will introduce eventId argument to the on()

doc/CustomEventConnector..md
```diff
@@ -0,0 +1,77 @@
...
+
+[Code](https://github.com/reshufflehq/reshuffle) |
+[npm](https://www.npmjs.com/package/reshuffle) |
+[Code sample](https://github.com/reshufflehq/reshuffle/examples/customEvent)
+
+
+### Reshuffle Custom Event Connector
+
+This package contains a [Reshuffle](https://github.com/reshufflehq/reshuffle)
+connector that can fire a customized event. Developer can fire it from a script.
```
duckranger, line 11: I think `custom` instead of `customized` is ok?
Also the last sentence can be better

doc/CustomEventConnector..md
```diff
@@ -0,0 +1,77 @@
...
+  )
+
+app.start()
+```
+
+### Table of Contents
+
+#### Connector Events
+
+[Fire Custom Events](#listen)
```
duckranger, line 43: I think 'Fire Custom Events' can be confusing as an event name. I was not sure whether this is an event or an action. Looking at the rest of the readme - I think the event and action were mixed into one...

duckranger (approved):
Yay!

PR #474, opened 2020-12-21
RS-209

lagache:
- add env variable to set an endpoint to get packages versions (RESHUFFLE_PKG_VERSION_PATH)
- change HEALTH_CHECK_PATH to RESHUFFLE_HEALTH_PATH
- reorder routes so generic routes goes at the very end

### DIFF
--- a/src/Reshuffle.ts
+++ b/src/Reshuffle.ts
@@ -6,16 +6,17 @@ import { BaseConnector, BaseHttpConnector, EventConfiguration } from 'reshuffle-
 import { createLogger } from './Logger'
 import { Logger, LoggerOptions } from 'winston'
 import http from 'http'
-import fs from 'fs'
+import { promises as fs } from 'fs'
 
-function getReshufflePackagesList(directory = __dirname): Record<string, string> {
+async function getReshufflePackagesList(directory = __dirname): Promise<Record<string, string>> {
   const filename = 'package.json'
   const dirSplit = directory.split('/')
   const dir = dirSplit.slice(0, dirSplit.length - 1).join('/')
 
   try {
-    if (fs.readdirSync(dir).includes(filename)) {
-      const packageJson = fs.readFileSync(`${dir}/${filename}`, 'utf8')
+    const listDir = await fs.readdir(dir)
+    if (listDir.includes(filename)) {
+      const packageJson = await fs.readFile(`${dir}/${filename}`, 'utf8')
       return JSON.parse(packageJson).dependencies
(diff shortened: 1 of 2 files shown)


src/Reshuffle.ts
```diff
@@ -6,6 +6,24 @@ import { BaseConnector, BaseHttpConnector, EventConfiguration } from 'reshuffle-
...
 import http from 'http'
+import fs from 'fs'
+
+function getReshufflePackagesList(directory = __dirname): Record<string, string> {
+  const filename = 'package.json'
+  const dirSplit = directory.split('/')
+  const dir = dirSplit.slice(0, dirSplit.length - 1).join('/')
+
+  try {
+    if (fs.readdirSync(dir).includes(filename)) {
```
avnerbraverman, line 17: readdirSync and readFileSync will block the entire event loop in NodeJS, which is not a good idea. Use "import { promises as fs } from 'fs'" and them "await readdir" and "await readFile"

src/Reshuffle.ts
```diff
@@ -6,6 +6,24 @@ import { BaseConnector, BaseHttpConnector, EventConfiguration } from 'reshuffle-
...
+      const packageJson = fs.readFileSync(`${dir}/${filename}`, 'utf8')
+      return JSON.parse(packageJson).dependencies
+    }
+
+    return getReshufflePackagesList(directory + '../')
```
avnerbraverman, line 22: Recursion. Bold!

avnerbraverman (changes requested):
LGTM. One change requested around async file access

src/Reshuffle.ts
```diff
@@ -6,6 +6,24 @@ import { BaseConnector, BaseHttpConnector, EventConfiguration } from 'reshuffle-
 import { createLogger } from './Logger'
 import { Logger, LoggerOptions } from 'winston'
 import http from 'http'
+import fs from 'fs'
+
+function getReshufflePackagesList(directory = __dirname): Record<string, string> {
```
duckranger, line 11: Suggestion: When I read the function name, I thought it would only be the `reshuffle` packages :) So maybe we can name it `getInstalledPackages` or something like this?

duckranger (approved):

src/Reshuffle.ts
lagache: very good point, fixed

avnerbraverman (approved):

ashevat (approved):

PR #477, opened 2020-12-23
custom-event-readme

zacmos:


src/Reshuffle.ts
```diff
@@ -6,6 +6,25 @@ import { BaseConnector, BaseHttpConnector, EventConfiguration } from 'reshuffle-
 import { createLogger } from './Logger'
 import { Logger, LoggerOptions } from 'winston'
 import http from 'http'
+import { promises as fs } from 'fs'
```
ashevat, line 9: @zacmos @duckranger why are changes in Reshuffle a part of this PR? Seems like a merge problem - please do not include these?

ashevat (changes requested):

PR #486, opened 2021-02-09
Improve HTTP Connector performance

duckranger:
The logging (a) did not contribute a thing and (b) slowed down soak tests by almost a 2x factor

### DIFF
--- a/src/connectors/HttpConnector.ts
+++ b/src/connectors/HttpConnector.ts
@@ -50,7 +50,6 @@ export default class HttpConnector extends BaseHttpConnector<
     )
 
     if (eventConfiguration) {
-      this.app.getLogger().info('Handling event')
 
       handled = await this.app.handleEvent(eventConfiguration.id, {
         ...eventConfiguration,


src/connectors/HttpConnector.ts
```diff
@@ -50,7 +50,6 @@ export default class HttpConnector extends BaseHttpConnector<
     )
 
     if (eventConfiguration) {
-      this.app.getLogger().info('Handling event')
```
ashevat, line 53: maybe replace with .debug() ?

duckranger: We could - but the log message itself adds nothing I think?

lagache: I think there's more to it in this story than removing the `handling event` message
I think it's relevant to keep it in Debug, or you never know if something happens.
I would do a few things on this story (there are suggestions only):
- add an option to EventConfiguration (silent true/false, default to false)
- each connector options take silent true/false
- if silent => no log at all, else keeping the current info and add the event.id (e.g. `Handling event Slack/message/123456789`))

ashevat (approved):

PR #488, opened 2021-02-19
improve types

refarer:


gabeeden:
Note: build needs https://github.com/reshufflehq/reshuffle-base-connector/pull/17 to be merged & published

src/Reshuffle.ts
```diff
@@ -74,7 +78,7 @@ export default class Reshuffle {
     }
   }
 
-  when(eventConfiguration: EventConfiguration, handler: (() => void) | Handler): Reshuffle {
+  when(eventConfiguration: EventConfiguration, handler: Handler | HandlerWrapper) {
```
lagache, line 81: do we have one connector using the HandlerWrapper? if no then we should just delete that type and the check i the code

lagache (approved):

src/Reshuffle.ts
gabeeden: I don't think so, just a test case. Let's remove it

PR #489, opened 2021-02-23
mysql-store

zacmos:


src/persistency/MySQLStoreAdapter.ts
```diff
@@ -0,0 +1,82 @@
...
+  public async del(key: string): Promise<void> {
+    await this.validateDB()
+    await this.pool.query(`DELETE FROM ${this.table} WHERE id = ?`, [key])
+  }
+
+  public async get(key: string): Promise<any> {
+    await this.validateDB()
+    const res = await this.pool.query(`SELECT value FROM ${this.table} WHERE id = ?`, [key])
+    const row = res[0][0]
+    return !row || row.length === 0 ? undefined : JSON.parse(row.value)
```
lagache, line 26: minor thing, you can check row using the ? and any positive number is truthy in JS and 0 is falsy so you can write it that way checking the length only:
This should do the same thing but shorter
`return row?.length ? JSON.parse(row.value) : undefined`

src/persistency/MySQLStoreAdapter.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  public async set(key: string, value: any): Promise<any> {
+    await this.validateDB()
+    const stringifyValue = JSON.stringify(value)
+    await this.pool.query(
+      `INSERT INTO ${this.table}(id, value) VALUES(?, ?) ` + 'ON DUPLICATE KEY UPDATE value = ?',
+      [key, stringifyValue, stringifyValue],
+    )
+    return value
```
lagache, line 43: a little strange to return the value here. Shall a setter returns something?

lagache (approved):

src/persistency/MySQLStoreAdapter.ts
zacmos: Same as SQLStoreAdapter

PR #491, opened 2021-03-03
mssql-store-adapter

zacmos:


package.json
```diff
@@ -30,12 +30,14 @@
...
   "dependencies": {
+    "@types/mssql": "^6.0.7",
+    "@types/node-cron": "^2.0.3",
     "async-mutex": "^0.2.4",
     "express": "^4.17.1",
     "is-valid-path": "^0.1.1",
+    "mssql": "^6.3.1",
     "nanoid": "^3.1.16",
     "node-cron": "^2.0.3",
-    "@types/node-cron": "^2.0.3",
```
duckranger, line 38: Why was this removed?

duckranger (changes requested):

package.json
zacmos: It's not removed but moved up in the file..

duckranger (approved):

PR #501, opened 2021-04-04
Allow custom middleware to be added to the express webserver

antonmarsden:
It's handy to be able to add middleware to the express webserver in some circumstances. A few examples:
- adding a health check URL (currently hardcoded into `Reshuffle.ts` but could be removed in future)
- adding in custom middleware such as a **`cookie-parser`**


lagache (approved):

lagache:
@antonmarsden thanks for your contribution, great addition to the codebase

